PR #139, opened 2016-02-17
Add: Domains

folletto:
From a discussion on Automattic/wp-calypso#3320, here's an idea for the Domains icon:

screen shot 2016-02-17 at 19 01 26

/cc @breezyskies @mikeshelton1503


drw158:
I think the idea of A or B is a viable solution, but in the end I think it's a bit cryptic. I don't think it's as good as the regular globe.

afsheenirani:
Just an idea...how about `url` with maybe a `+` sign for adding a domain? It might give you more pixel space than `www`.

folletto:
In general icons should try to be word-less, and I'm not sure "URL" is understandable to the general public enough to break that general idea. :)

Let's go for the globe @drw158 suggested then. :)

folletto:
screen shot 2016-03-02 at 11 52 26

Ok, here it is. :)

Are we all ok on this one for domains?

mikeshelton1503:
>Are we all ok on this one for domains?

:+1: I think this is a good compromise. I can't say it adds any clarity but domains is abstract so I don't think we're going to without being too complex.

breezyskies:
:+1:

Issue #98, opened 2015-09-18
Add script to generate `npm` package that can be used in Calypso


gcorne:

afsheenirani:
@retrofox, @jasmussen @gcorne - I think this is a great idea and I've been taking a look into it. I will submit a PR soon to kick-off discussions on the design based on a first draft that I've built. 

However, at a high level, to answer @retrofox's question...the way I'm envisioning it to work is that once you install and import the gridicons package into calypso, all you would need to do to display an icon should be:

`gridicons( 'iconName', iconSize, iconOnClickHandler )`

A call to the above function would end up returning the required `<svg></svg>` tag which can be easily embedded into any React components `render()` function. Given this, we may need to do away with the `Gridicons` React class in calypso (but that is a second step once we have a publicly available working npm gridicons package).

Here's what i think should be made available in the published npm package:

1. The index.js module which provides the `gridicons` function.
2. A LICSENSE.md file
3. A Readme.md file
4. Examples - This is something I would like to get a broader view on. Do we want to keep the examples + search of gridicons in the calypso project or should we just redirect this to the gridicons repo from calypso? My view is Calypso should redirect to the gridicons repo since the gridicons repo should be a wholistic project in itself.

In addition to what will be published we will need a few things for the generation and management of the npm package:

1. Updates to Gruntfile.js to generate the index.js module
2. A mechanism to test the generated index.js module locally on client side before publishing to npm.
 - I'm currently using Browserify to bundle up all `require` related node dependencies like `classnames` and expose this on the client side for testing. Do let me know if there are better ways to achieve this and I can explore that.
3. A shell script or Readme.md set of instructions on how to publish the npm package for newer releases.
 - Maintaining versioning of releases

cc: @folletto @drw158 @kellychoffman @nb  as well.

gwwar:
Should be resolved by #188

PR #145, opened 2016-03-14
Gridicons - NPM package Creation

afsheenirani:
## Background

This PR is to kick-off a discussion\design review of creating an npm package for the `gridicons` repository that can easily be installed and used in Calypso. 

At this point in time, the aim is to get a few folks to look at this proof-of-concept of how the gridicons npm package will be created, maintained, etc. The PR is certainly not in a state that it can be code reviewed, though code review comments are always welcome.

The auto-generation of the index.js files via grunt will be tackled in a future commit, once we finalize the approach of how/what features we want the npm module to export. However, I have committed the sample (with only sample icons) `index.js` files that "would" be what grunt would auto-generate for all icons.

Please note that the *.md files in this PR go into some detail of what the established process would be for maintaining and publishing the npm package.
## Reviewers

@retrofox @jasmussen @drw158 @kellychoffman @nb 
## Issue
#98


afsheenirani:
cc: @folletto, @aduth, @mtias for a design review before proceeding with the approach. I think the key points to review here are:
1. What is to be published? (i.e. contents of npm/publish/*)
2. Design/Interface of the `gridicons` API which will be exposed in npm in order to get a gridicon in any external project. Currently the focus here is to get a single gridicon, however, this feature set can be expanded in the future to provide more.
3. Workflow of how admins will test the new changes (i.e., client-side testing in browser - leveraging browserify to expose index.js to the browser).

npm/publish/index.js
```diff
@@ -0,0 +1,581 @@
...
+
+	return( path );
+};
+
+var gridicons = function( iconName, size, onClick ) {
+	var icon = 'gridicons-' + iconName,
+		needOffset = needsOffset( icon, size ),
+		needOffsetX = needsOffsetX( icon, size ),
+		needOffsetY = needsOffsetY( icon, size ),
+		svg, iconClass;
```
folletto: Can we use multiple `var` lines instead of the comma indented style that is harder to read?

i.e.

```
var icon = 'gridicons-' + iconName;
var needOffset = needsOffset( icon, size );
```

etc. :)

folletto:
In terms of documentation and code usability:

1. `PUBLISHING.md` — "Used grunt and generated a new npm/publish/index.js file." — is the task automated (i.e. run every time we run `grunt`)? If not, can we do that so there's no separate action to remember?
2. `PUBLISHING.md` — can we turn the readme in a numbered checklist first + in-depth blocks as needed later?
3. `index.js` — `function gridicons()` — does this allow optional parameters? Ideally I should be able to just write `gridicons( 'post' );` to get a default 24px icon with no `onClick` handler, as well as `gridicons( 'post', 18 );`.
4. `README.md` — contains a pseudo-bullet-list that is getting all "1." counting due to the list being broken in multiple code blocks. I'd try again to make it a short and to the point checklist with extra sections later if there's any need to clarify further.

folletto:
Aside: can't we just publish the React component itself following [these guidelines](http://react-components.com/)?

npm/publish/index.js
aduth: >Can we use multiple `var` lines instead of the comma indented style that is harder to read?

:-1: from me from both a personal preference and if we care to adhere to Calypso coding standards, for which this is a discouraged practice ([reference](https://github.com/Automattic/wp-calypso/blob/master/docs/coding-guidelines/javascript.md#declaring-variables-with-var)).

folletto: Ok, ok. I'm *very* strongly against that, but I won't argue here.

afsheenirani: @folletto / @aduth - Yes, I was just following coding standards. Also, the approach taken in coding standard makes sense to me because of `var` having the capability to trample it's values regardless of block scope.

afsheenirani:
@folletto -
1. Yes, `grunt` automates the generation of the file. Look at the changes to `Gruntfile.js`. The point I am making in `Publishing.md` is to ensure that before publishing, the admin should've run `grunt` to use the latest generated file.
2. I'll convert the `*.md` documentation to bulleted checklists with the more detailed explanation of the steps into later sections.
3. Good point on the optional parameters to `function gridicons()` in `index.js`. I'll have a look.
4. Will fix the Readme.md.

Lastly, but most importantly, we could look to publish the react component. I guess this is the piece I'd like to get more consensus on. The reason I went by the approach of publishing a `*.js` file is that it would be the "most common factor" thereby allowing for broader adoption e.g., if someone has an AngularJS project they could still `require` and use the `wp-gridicons` package.

Needless to say, we could additionally also publish a react component if that simplifies usage in Calypso. What I was envisioning is that we would have our own local React component defined in Calypso that will leverage the `wp-gridicons` package (if the need exists -- still TBD based on usages in Calypso).

folletto:
> The point I am making in Publishing.md is to ensure that before publishing, the admin should've run grunt to use the latest generated file.

Ok, if it's linked to grunt, that already has to be run as part of the merge process, so I wouldn't phrase it in a way that sounds like something different. I'd either take it out, or just mention something along the lines of "this is done during the normal grunt build step" so it's clear.

> The reason I went by the approach of publishing a *.js file is that it would be the "most common factor" thereby allowing for broader adoption e.g., if someone has an AngularJS project they could still require and use the `wp-gridicons` package.

Which, I'm on board, but:

1. There are some details that add complexity, such as `className` that is a React thing, and that can't be used if it's pure `js`, but also would mean some kind of workaround to make it work then inside React if that change is done.
2. I'd err on the side of addressing the immediate pragmatic need here, which is having a way to use Gridicons in Calypso using NPM.

afsheenirani:
@folletto / @aduth / @nb - I've made some enhancements, to summarize them:
1. I'm creating an additional API called `wpAllIcons` to provide a list of all the WP Gridicons. Let me know if we want to name this differently. Useful for people to programmatically explore the list of icons.
2. I've renamed the API by which we get a single gridicon to be `wpGridIcon` instead of `gridicons`, since there is no WP namespace and it is misleading since it gets you only one icon and not many icons.
3. As requested, I've added the capability to make the set of arguments to the `wpGridicon` optional (i.e., the Size and onClickHandler will have default values if the arguments are not passed in to the API.
4. Simplified and improved the `npm/test/index.js` file to simply `require` the `npm/publish/index.js` file for testing thereby removing code duplication and also putting browserify to better use.

I believe this should bring my changes into a "publishable" state. (There might be a few minor edits required in documentation to update `gridicons` to `wpGridicon`). However, unless there are any "broader" concerns, I'd like to proceed to publish this as a first patch version and then test importing into Calypso for usage.

(Lastly, I'm not sure if it's okay for me to publish to NPM on behalf of Automattic...)

afsheenirani:
@folletto / @aduth / @nb - Updated documentation to correctly point to `wp-gridicons` package and `wpGridicon` api.

drw158:
Is it possible to replicate what has been done for SocialLogos?

https://github.com/Automattic/social-logos

I believe Gridicons and SocialLogos are almost identical in terms of processes and structure. Whatever we do there should be ported over here.

cc @gwwar

gwwar:
> Is it possible to replicate what has been done for SocialLogos?

Yeah, definitely. I haven't had time to follow up on it, but I wanted to add tests/CI to social-logos so we could publish to npm more easily without having to sync up PRs with Calypso.

If folks want to take a stab at this, also see:

https://github.com/Automattic/social-logos/pull/41
https://github.com/Automattic/social-logos/pull/42
https://github.com/Automattic/social-logos/pull/43

package.json
```diff
@@ -2,6 +2,7 @@
   "name": "gridicons",
   "version": "0.1.0",
   "devDependencies": {
+    "classnames": "^2.2.3",
```
gwwar: Careful here. This would break in a prod install. I would recommend removing the classnames dep entirely.

gwwar:
Any reason for the npm folder namespace? That's rather unusual.

drw158:
now obsolete because of #188 ?

gwwar:
Closing for #188 instead

PR #148, opened 2016-03-21
Remove backspace icon

### LINKED ISSUE #146: Remove the 'backspace' Gridicon
This icon has been in the repo since the beginning, but I don't think I've ever seen it used anywhere. The style of Gridicons has changed, and I think that it either needs to be removed, or tweaked to match the rest of the set.

screen shot 2016-03-17 at 10 56 09 am

drw158:
fixes #146


drw158:
@jasmussen could you do a quick review if you have a minute? I know it's out of the ordinary, since we are removing an icon...

I basically search for all instances of `backspace`, but `grunt` took take of a lot of the work.

jasmussen:
There was a trace of backspace in a react component, it needed offset when shown at 18px. I removed that and grunted. Looks good to me! :+1:

PR #150, opened 2016-03-22
Update plus icon

drw158:
Before, left
After, right

screen shot 2016-03-22 at 4 14 02 pm

The current plus icon appears too small in relation with the other icons.

Cross icon is updated also, for the same reasons:

screen shot 2016-03-22 at 4 32 23 pm

Thanks to @mattmiklic for pointing this out


drw158:
I will go ahead and add to .ai file and generate.

drw158:
@jasmussen For some reason, running the pdf script is adding a lot of pdf files, even thought they already exist. I'm not sure what is going on. Some of the changes are just bumping the version in the pdf file.

Check out commit https://github.com/Automattic/gridicons/pull/150/commits/4e6f0f4f11fe0393ab0cbf64b7331a52ae5c6449

Other than that, I think this PR is ready.

drw158:
It's like the script renamed all of the pdf files, but the filenames are the same...

jasmussen:
> For some reason, running the pdf script is adding a lot of pdf files, even thought they already exist. I'm not sure what is going on. Some of the changes are just bumping the version in the pdf file.

Took a look, and things are looking good to me. 

What happened was that your version of `cairosvg` was newer than mine by one point, and apparently Cairo outputs its version number under a "generator" tag in the header of the PDF file. Since your version was newer than the one I used to generate all the initial PDFs, every PDF was technically different, even though the only difference was a bumped version number.

I don't know that there's anything we can do about this. Doesn't look like there's an option in Cairo to not bake in the version number in the PDFs. So I think we'll just have to live with this. But thanks for checking in, I think I'll add a note to the readme.

As a sidebar, the design changes to the plus look good, here's a :+1:

drw158:
Thanks!

PR #152, opened 2016-04-14
Add icon: Reader Follow and Following

jancavan:
Add Reader Follow and Following icons:

gridicons-follow-following

/cc @drw158 @jasmussen 

Related: https://github.com/Automattic/wp-calypso/issues/3121


jasmussen:
Looks great to me!

Do we need also a version with a minus overlaid, for when you mean to unfollow something?

drw158:
Removed some unnecessary points

screen shot 2016-04-15 at 11 15 43 am

I'd like to get the check mark on some even grid numbers, but that can wait for another PR I think.

screen shot 2016-04-15 at 11 17 39 am

I worry about how top-left heavy these icons are, and how they look with the other icons. They might not appear center. We can fix this later.

Other than that, looking good. Do you need help adding the icons and running grunt?

jancavan:
Thanks for your feedback @jasmussen and @drw158!

> Do we need also a version with a minus overlaid, for when you mean to unfollow something?

Yup, but we'll just add this in later. I'd like to roll this out soon so we can also proceed with testing.

> Removed some unnecessary points

Thanks for cleaning this up!

> I'd like to get the check mark on some even grid numbers

I'm using this 2px [checkmark icon](https://github.com/Automattic/gridicons/blob/master/svg/gridicons-checkmark.svg), but a tweaked/smaller version. I made the gap 2px as well, but this does cause the bottom of the checkmark not aligning with the grid. Any ideas?

screenshot-2016-04-15-10 23 19

jancavan:
On https://github.com/Automattic/gridicons/pull/152/commits/12be2d1a6d39c80a526c0241e26262a1905b61d7, I've made the plus and check symbols slightly smaller. 

The plus now matches the [add image gridicon](https://github.com/Automattic/gridicons/blob/master/svg/gridicons-add-image.svg) plus.

gridicons

> Do you need help adding the icons and running grunt?

@drw158 Yup, not sure what next steps are after review. Is that something you can help me with?

drw158:
I noticed that adding a 2px stroke to the check has some misalignment with the reader icon. Looking into it now.

screen shot 2016-04-15 at 4 10 10 pm

drw158:
Actually, I don't have time to look into it right now. I think the new icons you made with the smaller icons look good. Try adding a stroke to the new check, I think there is still some misalignment. 

Also, the check vector is an open shape, just close it up.

I can help with the next steps. I can do it for you, or you can try it yourself. Ping me on Slack, and I can walk you through it if the instructions aren't clear. Follow the "Add a Proposed Icon to Gridicons (Admins Only" part in the readme.

jancavan:
Thanks @drw158. Fixed those on https://github.com/Automattic/gridicons/pull/152/commits/7bab3423ec8820be7b5c0a21e769efa5462055ae.

I'm going to try adding the icons to Gridicons.

jancavan:
This is ready for review.

/cc @jasmussen @drw158

drw158:
Can you ungroup the icons, then select the icon and do a `Make` > `Compound Path`? These 2 icons should just show a single compound path in the layer.

screen shot 2016-04-18 at 3 13 30 pm

Maybe it's just me, but adding a 2pt stroke to the check is still not aligned. But, it probably doesn't really matter that much, so I don't know.

screen shot 2016-04-18 at 3 11 36 pm

Lastly, there shouldn't be any changes to the icon-template file.

jancavan:
That's very odd, not sure what's going on but the icon I created seems be aligned, but it creates that misalignment in the svgs:


screenshot-2016-04-18-13 31 56

Let me take a look again.

jancavan:
@drw158 @jasmussen reverted back `icon-template.ai` file to that version in Master (https://github.com/Automattic/gridicons/pull/152/commits/230d5e163a4d8c3eeb3128d8ebd6f21c98567c1e) and fixed misalignment and converted to Compound Path (https://github.com/Automattic/gridicons/pull/152/commits/955abb22686b9450b07844192249c8e52d06eff3).

Is this good to?

jasmussen:
Looks pretty good to me!

jancavan:
@jasmussen @drw158 I just found this out today, but I'm having an issue for when any of the Reader icons is less than `24x24 px` in size. It ends up looking blurry:
screenshot 2016-05-06 14 37 51

I initially thought that this may be because the Follow/Following icons don't have equal surrounding margins, but I tested the base Reader icon and it's also blurry:

screenshot 2016-05-06 14 39 41

But looking at their pixel preview in Illustrator, I don't see any particular cause as to why this is happening:
pixel

Any ideas?

jasmussen:
The sad truth you've discovered here, is that unfortunately this is to be expected.

Right now we have one icon set, that's drawn to look best at 24px. When you scale that up 2x or 3x or 5x, it looks good too. If you scale it down to exactly .5x, it also looks good, albeit very small at 12px.

18px is .75x, which in many cases looks pretty good (circles, simple shapes), in other cases it'll still look blurry (horizontal or vertical lines).

Specifically in this use case, I'd try the 24px icon. I think it'll look good.

jancavan:
Thanks @joen. I think we can manage to use `24px` all throughout Reader, but the issue I was having was with the Author module on Stats, which has to be smaller than `24px`. This one is sized `18px` like most of the Stats modules icons:

screenshot 2016-05-10 09 56 49

jasmussen:
Yep, that's a challenge, and lots of duplicate text and buttons. I wish there was some way we could reduce that UI, but I can't think of one.

In any case, the 18px size blur is unavoidable, unless we draw new versions just for that size, which is a challenge in its own right. 

Could the follow button work without a text label, icon only? And in such a situation could we use the 24px version?

jancavan:
> Could the follow button work without a text label, icon only?

Yup, although we only do this in viewports `<480px`. That's in Reader though, not in Stats.

> And in such a situation could we use the 24px version?

For the Authors module in Stats, `24px` is still big - but that's also relative to the rest of the icons. I think it could work, but the rest of the icons' size would need to be updated so they all match.

jasmussen:
Gotcha, thanks. Then we're left with accepting this blur. I think it's forgivable. I will try our offset-hack as well, that might make them slightly crisper.

jasmussen:
Hey @jancavan, I tried applying the "offset-hack" i keep talking about. What it does is nudge the icon on the sub-pixel level, which in some cases makes it look more crisp. Here are the results:

No offset-hack (how it looks now):

no-offset

Offset on both x and y axes: 

offset-x-y

Offset on only the x axis:

offset-x

I think the version that offsets both axes looks pretty decent. Let me know what you think and I'll add that to the offset-array.

jancavan:
@jasmussen albeit small, I do notice a difference. Thanks for doing this. Is there anything I need to do on my end?

jasmussen:
>  Is there anything I need to do on my end?

Nothing but pick which one of the three you like most :) if you don't pick, I'll go with the 2nd option (both axes offset).

jancavan:
> Offset on both x and y axes

This one looks better. Thanks :)

PR #153, opened 2016-05-06
Add "More" icon

jasmussen:
This is an icon for the format bar. Specifically it is to insert the `<!--more-->` tag.

Here's the equivalent Dashicon we're replacing:

screen shot 2016-05-06 at 09 46 54

Here's the Gridicon version:

screen shot 2016-05-06 at 09 49 08


drw158:
Naming it "more" might be a concern with the ellipsis, and confusing the two since we regularly mention the ellipsis icon as the "more" icon.  How about 'read-more'?

jasmussen:
Excellent point, I'll rename, thanks!

drw158:
LGTM 👍

PR #154, opened 2016-05-12
Add offset x and y to react component.

jasmussen:
Per discussion on the new Reader follow and following icons, this PR adds those to the offset array to look slightly crisper at 18 and 36px sizes.


drw158:
this looks good to me

Issue #144, opened 2016-03-08
Custom letterforms for icons (no Open Sans)


drw158:
I think we should consider custom letterforms, for some of our icons. The ? icon especially could use some help.

Left is current, options follow

![screen shot 2016-02-17 at 3 48 58 pm](https://cloud.githubusercontent.com/assets/618551/13619676/1bc26338-e550-11e5-8c77-f9b455fa60f1.png)

mattmiklic:
I like the second from left, though it feels like it needs at least a little bit of a stem descending from the aperture of the `?`. Definitely feels like a better match for the icon style than the Open Sans-based glyph.

jasmussen:
I also dig the second one from the left. 

As a sidenote, as small as this sounds, it'd be good to think up some guidelines for what drives these custom letterforms, outside of just the general Gridicon guidelines. I noticed Google added a bunch of custom letterforms to their material set recently, and it's quickly clear there's a ruleset to their letterforms:

![screen shot 2016-03-09 at 11 50 11](https://cloud.githubusercontent.com/assets/1204802/13633096/601c6cc0-e5ed-11e5-88db-4b09afb564fb.png)

![screen shot 2016-03-09 at 11 50 15](https://cloud.githubusercontent.com/assets/1204802/13633099/63070530-e5ed-11e5-8c5a-a86f22cd6e3a.png)

![screen shot 2016-03-09 at 11 50 36](https://cloud.githubusercontent.com/assets/1204802/13633104/650559a4-e5ed-11e5-9971-f05cefa370aa.png)

![screen shot 2016-03-09 at 11 50 41](https://cloud.githubusercontent.com/assets/1204802/13633105/670ddc8a-e5ed-11e5-8a2a-e1cdca788188.png)

The material style here is a little bit machinist and brutalist for my tastes. But I like how it's clearly a font, while at the same time clearly using the icon design language.

drw158:
Some more options

![screen shot 2016-03-21 at 3 01 30 pm](https://cloud.githubusercontent.com/assets/618551/13932425/fd66e7bc-ef75-11e5-83de-5f61aad5552c.png)

jasmussen:
1 and 3 look amazing, and 3 is my favorite. Nice work!

mattmiklic:
Agreed, I like both 1 and 3; slight aesthetic preference for 3, but if we want these to adhere to the "straight 45º angles" rule then I'd go with 1.

drw158:
[...]

PR #156, opened 2016-05-19
New updates to help icons

drw158:
## Before/After

screen shot 2016-05-19 at 3 42 01 pm
screen shot 2016-05-19 at 3 41 32 pm


drw158:
fixes #144

melchoyce:
New one is much better.

rickybanister:
the only downside I see to the new icon is that at small size the dot is very close to the rest of the figure. The old one had a little more breathing room, is that fixable?

jasmussen:
What Rick said, otherwise 👍

folletto:
👍

drw158:
Yea, I tried to keep that space 2px, but I couldn't work it out. It would make the ? too tall, and therefore not centered. Also, there wouldn't be room for the outlined version.

screen shot 2016-05-20 at 10 15 54 am

I was encouraged though, that the Material help icons have a 1px space on the outlined version as well:

screen shot 2016-05-20 at 10 13 31 am

jasmussen:
The original is nice enough to merge, I think, so all's good on that front. However I can't help but ask: do we actually need an outline version?

folletto:
>  However I can't help but ask: do we actually need an outline version?

If we have a rationale, I think it's fine...

...

...do we have a clear way to tell when to use one or the other? :D

drw158:
Here's more options 

screen shot 2016-05-20 at 10 45 20 am

drw158:
Outline version is there as an option for designers.

drw158:
I'm gonna go ahead and merge this because of https://github.com/Automattic/gridicons/pull/156#issuecomment-220641017

As always, it can be iterated on in the future.

PR #157, opened 2016-05-26
New icon: Coupon

jameskoster:
Adds the coupon icon.

Coupon


folletto:
Maybe it's just me, but... doesn't look more like a razor blade? (Ignore if it's just me)
A dotted line on a side might help. 

Also it seems to me that there are too many line weight. Might be an issue related with dealing with the edge cuts, but maybe we can deal with that? Just removing the inner rectangle might help (would take away the "inner transparent line" that adds to it.

jameskoster:
Now that you mention it, yes. It does look a little like a razor blade.

I have this as a draft which might be more coupon'y?

coupon

folletto:
I have to say that I have no clear mental image of how a coupon looks like, so I'm a bit at loss here to give an idea on a general direction. The general direction of the first seemed ok, also because it was more unique, the above could be associated with a general post for example.

jameskoster:
No inner rectangle; 

coupon

folletto:
To me that feels the best of the three. :)

jameskoster:
I'll update the PR and await further feedback xD

kellychoffman:
Agree, the last one with no inner rectangle looks the best.

drw158:
Nice. Yea, coupon seems really hard to design for. I think this works, although I would name it `ticket`. If we ever need a `ticket` and `coupon` icon in the future, then we'll have to differentiate the two.

One idea for coupon:

screen shot 2016-05-26 at 10 35 53 am

Also, we rarely tilt icons at an angle, and I don't think this one requires it. Here are some ideas:

screen shot 2016-05-26 at 10 51 47 am

jameskoster:
I like 1 or 2 :)

drw158:
Updated icon to number 2:

screen shot 2016-05-27 at 2 54 54 pm

jameskoster:
I like the icon.

Part of me is still wondering if we can create something better for coupons specifically. The proposed icon is maybe more 'ticket' than 'coupon'.

I think I'll have one more stab at an alternative (or an addition, a ticket icon would probably be useful anyway right?). Will share asap :)

folletto:
Brainstorming (the proportions and sizing are all off, it's really just to throw ideas in), borrowing from a few of your concepts above:

screen shot 2016-05-31 at 12 08 01

jameskoster:
Yeah I thought about an outline/dotted line. The first is pretty nice.

Here's my idea for a more traditional, cut-out coupon;

coupon

drw158:
idea:

screen shot 2016-05-31 at 5 32 24 pm

jameskoster:
not sure the scissors really stand up without the holes in the handle. They could be knitting needles, swords, pipettes. Maybe it's just the angle (or just me). 

I think we definitely need the dotted border.

If the scissors were tidied up a bit, having them inside the box would be fine I think.

folletto:
I'd still prefer the [one before](https://camo.githubusercontent.com/d9b38e4bda79d2378bacf9d91c78ebdd72b39bb8/68747470733a2f2f636c6475702e636f6d2f6b5137666b766e6e6d422e7468756d622e706e67), but I agree that with some shape refinement this could work too :)

drw158:
options:

screen shot 2016-06-01 at 2 39 23 pm
screen shot 2016-06-01 at 2 39 47 pm

jameskoster:
first one is looking pretty good to me!

folletto:
First one for me works as well. 👍

drw158:
updated 

screen shot 2016-06-02 at 10 43 39 am
screen shot 2016-06-02 at 10 43 20 am

PR #158, opened 2016-05-26
New icon: Credit card verification

jameskoster:
ccv

### DIFF
--- a/svg/gridicons-credit-cart-verification.svg
+++ b/svg/gridicons-credit-cart-verification.svg
@@ -1,0 +1,22 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Generator: Adobe Illustrator 19.1.0, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->
+<svg version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px"
+	 viewBox="0 0 24 24" enable-background="new 0 0 24 24" xml:space="preserve">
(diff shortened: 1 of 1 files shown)


folletto:
I feel there are a lot of different line weight in here, also the corner cut out in that way is a little too thin on smaller sizes. Can we normalize weights and simplify the shape?

jameskoster:
Yeah I guess the cut-out isn't necessary. How do you feel about this;

ccv)

folletto:
Definitely better.

A clarification: "verification" here means "to check the verification number" or "card has been verified"? I'm asking because if it's the second then a ✓ might work better in terms of symbology.

jameskoster:
It's an icon we use to illustrate the location of a cards security code (last 3 digits of number on the back).

folletto:
Ok it's perfect then. :)

drw158:
Nice, at first I liked the idea of using the checkmark with the credit card icon. I would consider just using the card icon with the code shown, like this:

screen shot 2016-05-26 at 10 28 57 am

Or maybe just delete the left portion of the number on your icon, like so:

screen shot 2016-05-26 at 10 32 17 am

Either way, I think the latest iteration on your icon works.

jameskoster:
I think having the left portion is helpful as it's quite common for the last 3 digits of a longer number to be requested. This helps demonstrate that.

I'll double check that we definitely still need this icon. Might be one we can skip.

jameskoster:
Just confirmed, this is only used on the frontend. We'll create a separate dedicated svg but it doesn't need to go in gridicons.

Issue #31, opened 2015-05-28
Add icons: Follow and Following


davewhitley:
My initial idea was to make the "follow" and "following" icons a square, since we now make the distinction between squares and circles for blavatars and avatars:

![screen shot 2015-05-28 at 4 05 42 pm](https://cloud.githubusercontent.com/assets/618551/7871133/9778e82c-0553-11e5-84d9-093568d7ff45.png)

Another option is to use a circle instead, which looks good, but doesn't have the nice the square vs. circle logic.

![screen shot 2015-05-28 at 4 11 47 pm](https://cloud.githubusercontent.com/assets/618551/7871251/52531adc-0554-11e5-9166-3c25a3bc8245.png)

Also, it might conflict with other uses of the plus in a circle icon:

![screen shot 2015-05-28 at 4 12 41 pm](https://cloud.githubusercontent.com/assets/618551/7871266/704e1082-0554-11e5-829f-ab47a5661000.png)

PR #159, opened 2016-05-26
New icon: Order cancelled

jameskoster:
outline
solid


folletto:
I would try to keep here the same proportions as the `+` icon, so a smaller `x`:

jameskoster:
cancelled
cancelled

folletto:
👍

drw158:
Just style variants, or do they mean 2 different things?

I would think about using the `block` icon instead. Maybe it depends on the interface these will be used in. Either way, these are pretty generic, so I'm not opposed. I'm worried these might be used for "close". That might be ok though.

folletto:
>  I'm worried these might be used for "close". That might be ok though.

Good point actually.

Do you think we should have an icon for "order"?

jameskoster:
> I'm worried these might be used for "close". That might be ok though.

I thought that'd be a benefit? It's true that might be a more frequent use-case though so perhaps we should use that for the name?

Not sure if I should open a separate issue for this, but I was originally planning to use `checkmark-circle` for completed orders. But there's not an 'outline' counterpart. Worth creating a separate icon for completed orders?

> Do you think we should have an icon for "order"?

Would probably be useful.

drw158:
> I thought that'd be a benefit? It's true that might be a more frequent use-case though so perhaps we should use that for the name?

I don't imagine the overlap in usage being a big problem; the designer of the interface will just need to make sure they don't use them together

> Not sure if I should open a separate issue for this, but I was originally planning to use checkmark-circle for completed orders.

We aren't using that icon for the Reader anymore, so maybe it'd be worth spending some time on.

Imo, this might be the best way to go about it:

- don't worry about making outline versions (not necessary)
- don't make an `order` icon

screen shot 2016-05-26 at 11 20 15 am
screen shot 2016-05-26 at 11 22 29 am


order complete, canceled, paused, and processing.

folletto:
The first 4 looks like a good combo to me. :)

drw158:
Or, outlined version: 

screen shot 2016-05-26 at 11 27 22 am

jameskoster:
> The first 4 looks like a good combo to me. :)

I agree. I think we should be consistent and make them all have the containing circle.

Speaking of consistent, should we update `checkmark-circle` so that the check mark isn't 'breaking out' of the circle?

drw158:
Here's the discussion on the check mark circle https://github.com/Automattic/gridicons/issues/31#issuecomment-121731004

I do think it has more character with it breaking out. I'm on the fence.

Could do something like this:

screen shot 2016-05-26 at 12 28 18 pm

folletto:
> I do think it has more character with it breaking out.

Agreed. Adds definitely character to it, and you mock nails both styles imho. :)

drw158:
I propose:

- Go with the first row here: https://github.com/Automattic/gridicons/pull/159#issuecomment-221921747
- Don't do the outlined version of these quite yet.
- I would name this icon `gridicons-cross-circle`

screen shot 2016-05-27 at 3 00 06 pm

folletto:
👍  makes sense

drw158:
Updated as outlined above

PR #161, opened 2016-05-26
New icon: Order processing

jameskoster:
outline
solid


folletto:
We might want to give the same "spacing" as mentioned in #159, like `+`:

jameskoster:
I think the only way we'd be able to do that would be to have the dots not fit the grid. Otherwise I think they're too small?

processing

folletto:
screen shot 2016-05-26 at 15 31 21

Yeah, the dots might be a bit bigger, to be optically of the same size of the lines. :)

drw158:
I like the idea. I would try using 1.5px dots — that is a trick that we can use when we run into these situations.

jameskoster:
The one on the left has 1.5px dots.

processing

drw158:
I think this will work.

drw158:
Maybe `gridicons-more-circle` would be more generic and appropriate

drw158:
Removed outlined icon and renamed main icon to `gridicons-ellipsis-circle`

Issue #47, opened 2015-07-15
Add icon: Price


drw158:
Previously mentioned here: #35

drw158:
The work I've done on price so far (ignore the stray dots)

![screen shot 2016-05-26 at 9 49 52 am](https://cloud.githubusercontent.com/assets/618551/15578733/4598d33a-2327-11e6-984e-16a7a669c8f9.png)

folletto:
This seems a good direction:

![screen shot 2016-05-26 at 15 52 56](https://cloud.githubusercontent.com/assets/4389/15578828/f1881d72-2359-11e6-9d17-d1201f5bcbbd.png)

Would be worth adding a slight contour to the dot to make a face silhouette? :D

drw158:
![screen shot 2016-05-27 at 3 08 17 pm](https://cloud.githubusercontent.com/assets/618551/15619928/dd569ba8-241c-11e6-8c4b-79cb759bbc1e.png)
![screen shot 2016-05-27 at 3 08 08 pm](https://cloud.githubusercontent.com/assets/618551/15619930/df67e8b6-241c-11e6-8c03-ca14d5a78b73.png)

PR #162, opened 2016-05-26
New icon: Order refunded

jameskoster:
outline
solid


folletto:
We might want to give the same "spacing" as mentioned in #159, like `+`:

Even if here might be harder given the shape of the arrow. Maybe instead of a "U" turn could be just a "L"?

jameskoster:
refund
refund

jameskoster:
Hmm, I'm not sure this is such a great metaphor for a refund actually.

Maybe the arrow should be pointing backwards?

folletto:
I agree. The latest one is good in terms of alignment with our patterns, but meaning is a little less evident.

What about arrow down pointing a coin?

drw158:
I think I would prefer just to use our `undo` icon for this instead, unless we are also using `undo` elsewhere in the same interface. I think using existing icons is usually better. 

Here are some ideas if we decide to create a new icon:

screen shot 2016-05-26 at 10 19 01 am
screen shot 2016-05-26 at 10 18 29 am

Maybe we relate it to whatever we do for the price icon? #47

folletto:
> Maybe we relate it to whatever we do for the price icon?

Great idea. That + arrow turning back (first icon above) could work well :)

drw158:
Idea for refund now that we have a price icon:

screen shot 2016-06-06 at 11 20 53 am
screen shot 2016-06-06 at 11 20 38 am

I still think we could just use the `undo` icon for refund.

folletto:
Have you tried also the bill inside the circular arrow?

This looks ok, but I'd reverse the direction of the arrow. ;)

drw158:
Yep makes more sense with the arrow reversed :)

screen shot 2016-06-06 at 11 29 13 am

I can't seem to put the price icon within an arrow circle and stick to our standards:

screen shot 2016-06-06 at 11 38 46 am

But I had the idea of using the rotate icon (circle face used because of space concerns):

screen shot 2016-06-06 at 11 46 13 am

screen shot 2016-06-06 at 11 46 21 am

folletto:
Agreed, the circle is a bit difficult with our standards.

The arrow up is good, and I agree the last one is the best one. That's the closest one to the meaning we want to express. 👍

drw158:
Ok, thanks Davide, I will proceed with this one.

> The arrow up is good, and I agree the last one is the best one. That's the closest one to the meaning we want to express. 👍

PR #164, opened 2016-05-26
New icon: Product downloadable

jameskoster:
product downloadable


folletto:
Can we use our standard arrow shape?

i.e. 

screen shot 2016-05-26 at 14 25 43

jameskoster:
Heh, I grabbed this arrow from https://cloudup.com/cYKtx6YCYjt

But yeah, that one would probably look better. Will update.

folletto:
Now that you point it out... that's odd too :D

jameskoster:
Looks better 👍 

download

drw158:
Nice, this idea works well.

drw158:
Updated 

screen shot 2016-05-31 at 4 26 53 pm

jameskoster:
The other product icons have a 2 pixel gap between the edge of the box and the icon within. We could do the same with this if we shrunk the 'shaft' of the arrow by 2 px. Worth it?

drw158:
When viewed at 100% I feel it's more optically balanced than if it had a 2px gap.

screen shot 2016-06-01 at 11 48 38 am

Imo, the arrow on the left feels like it's too far down, the right arrow feels centered and even.

folletto:
I agree. The right, longer one looks better balanced.

PR #165, opened 2016-05-26
New icon: Product grouped

jameskoster:
product grouped


folletto:
Nice.

I'd probably cut straight the bottom line of the box behind, instead of trying to contour the other box outline.

jameskoster:
good call, updated.

drw158:
Updated 

screen shot 2016-05-31 at 5 14 45 pm

jameskoster:
We may not be using this icon now so this can be put on ice. Not sure if it's worth creating a label to that effect?

PR #166, opened 2016-05-26
New icon: Product variable

jameskoster:
product variable

We've had various versions of this icon since WooCommerce 1.0 and have never really come up with something that _feels_ right. I've played on the 'types' icon that already exists in Gridicons but if anyone has a better idea I'm all ears :)

### DIFF
--- a/svg/gridicons-product-variable.svg
+++ b/svg/gridicons-product-variable.svg
@@ -1,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Generator: Adobe Illustrator 19.1.0, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->
+<svg version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px"
+	 viewBox="0 0 24 24" enable-background="new 0 0 24 24" xml:space="preserve">
+<g id="Adv._Guides">
+</g>
+<g id="Guides">
+</g>
+<g id="Artwork">
+	<g id="Layer_1_1_">
+	</g>
+	<g id="Layer_1_2_">
+	</g>
+	<g>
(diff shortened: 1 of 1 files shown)


folletto:
The way they are cut feels a bit odd. Does it get too small if we just use a scaled down and pixel-tweaked version of `gridicons-types`?

jameskoster:
Doesn't look great imo;

variable

folletto:
Agreed. We need a third way. :D

drw158:
What does this icon represent? I'm not sure what product variable means.

jameskoster:
@drw158 A variable product would be something like a tshirt. A single product (cotton v-neck) available in different color/size combinations. Each product variation is treated as a unique product with it's own sku.

drw158:
Maybe we could use the CPT icon for this

screen shot 2016-05-31 at 5 06 53 pm

jameskoster:
I think it should follow the format of the other product icons, personally. These icons are all displayed in close proximity, it'd look a little odd to have one so different I think.

folletto:
> I think it should follow the format of the other product icons, personally. 

Agreed. 👍

drw158:
Ok, I'll give it a go.

drw158:
This one has been the toughest. I don't think using the "types" icon in some way is going to work. It's a lot of information to squeeze.

Aside: I think "variation" might be the better word to use here. Not sure how this is presented in the Woo UI, but variable is less easy to grok than variation imo.

Ideas:

Here are some interesting concepts I found:

screen shot 2016-06-16 at 5 01 08 pm
screen shot 2016-06-16 at 5 00 59 pm

Maybe something like

screen shot 2016-06-16 at 5 04 36 pm

jameskoster:
Yeah this one is a challenge!

Hmmm, I don't know. That doesn't feel quite right to me and the balance seems odd. 

Perhaps we can play on one of the most popular variations folks use; color. Have two products 'layered' but fill the front product so that it appears black with the one behind being white. 

rough example;

variable

jameskoster:
We may not be using this icon now so this can be put on ice. Not sure if it's worth creating a label to that effect?

folletto:
I'd probably just close the issue(s). The code stays there, and we can always to search for it. Or do the designs again.

PR #167, opened 2016-05-26
New icon: Product virtual

jameskoster:
product virtual

PR #163, opened 2016-05-26
New icon: Product external

jameskoster:
external product

Issue #151, opened 2016-03-31
Add icon: package


kellychoffman:
Looking for a box/package icon for shipping.

Something like 

<img width="88" alt="screen shot 2016-03-31 at 10 27 28 am" src="https://cloud.githubusercontent.com/assets/5835847/14178996/3bc4e538-f72b-11e5-87f5-b2194d95c2eb.png">
<img width="80" alt="screen shot 2016-03-31 at 10 27 32 am" src="https://cloud.githubusercontent.com/assets/5835847/14178995/3bc3a1f0-f72b-11e5-9fa4-705d4bfcf26c.png">
<img width="90" alt="screen shot 2016-03-31 at 10 27 39 am" src="https://cloud.githubusercontent.com/assets/5835847/14178994/3bc184ce-f72b-11e5-8678-8942caff50ae.png">

drw158:
Since we want to keep everything 2d, this will be hard, but here are some icons I whipped up for "shipment"

![screen shot 2016-03-31 at 12 09 44 pm](https://cloud.githubusercontent.com/assets/618551/14184338/4d089bbe-f73a-11e5-8fba-3124af4c2245.png)

kellychoffman:
Nice. The truck for 'shipping' is great. For 'package' I think the first or third one, but I'm not sure how the third one would look small. I think a single package would be ideal because one of the ways I want to use it is to denote the package type: a box versus an envelope.

drw158:
Cool, I think the first one will work well.

kellychoffman:
Agree! Thanks, Dave.

PR #168, opened 2016-05-26
New icon: Product

jameskoster:
product

### DIFF
--- a/svg/gridicons-product.svg
+++ b/svg/gridicons-product.svg
@@ -1,0 +1,16 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Generator: Adobe Illustrator 19.1.0, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->
+<svg version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px"
+	 viewBox="0 0 24 24" enable-background="new 0 0 24 24" xml:space="preserve">
+<g id="Adv._Guides">
+</g>
+<g id="Guides">
+</g>
+<g id="Artwork">
+	<g id="Layer_1_1_">
+	</g>
+	<g id="Layer_1_2_">
+	</g>
+	<path d="M20,2H4L2,6v16h20V6L20,2z M18.8,4l0.5,1H4.7l0.5-1H18.8z M4,20V7h16v13H4z"/>
+</g>
+</svg>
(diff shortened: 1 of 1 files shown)


folletto:
👍  in line with the others. :)

kellychoffman:
👍  Looks good. We can use the for 'package' too instead of: https://github.com/Automattic/gridicons/issues/151

drw158:
Also not two dimensional. Here are some ideas:

screen shot 2016-05-26 at 10 02 22 am

screen shot 2016-05-26 at 10 05 09 am

folletto:
...it's 2d if you imagine it being a box with a lid partially open :P 

But I get the point, yes. To me this is a doable (and contained) exception, but I'd prefer too to go fully 2d as expected.

The pure square is too generic to be used here imho, and we need to consider all the other usages.

@jameskoster can you post a screenshot here with all the icons so far that use the "box" shape?

folletto:
Also cross referencing all the related issues: #163, #164, #165, #166, #167, #169.

jameskoster:
Current box icons;

boxes

folletto:
Ok some brainstorming:

screen shot 2016-05-26 at 16 30 03

jameskoster:
I think the first is the only feasible one seeing as we need to put something 'inside' the box to demonstrate the type of product? There's probably not enough room to do that with the other concepts.

drw158:
Here's a bunch of ideas going off of @folletto. Sorry it's messy.

screen shot 2016-05-26 at 12 16 31 pm

jameskoster:
How about this @drw158 @folletto ?

products

It gives an exact square for the main box area which is nice for putting things inside. Needs some tidying but you get the idea.

folletto:
That idea seems working well to me! :)

jameskoster:
Tidied up a bit;

icons

drw158:
I think you are headed in the right direction. I think we could give a bit more personality to the regular package. The proportions of a perfect square don't sit quite well to me, and your icons might seem slightly larger compared to the other icons in Gridicons. Here is my take:

screen shot 2016-05-27 at 10 03 12 am
screen shot 2016-05-27 at 10 05 30 am

We should simplify the multiple package icon; I think it should be similar to the multiple images one. It's hard to make it not look awkward in such a small space:

screen shot 2016-05-27 at 10 07 50 am
screen shot 2016-05-27 at 10 07 38 am

folletto:
That looks great :)

jameskoster:
Yup that looks like a winner! Would you mind updating the relevant PR's @drw158 ?

drw158:
Updated product icon

screen shot 2016-05-31 at 3 54 27 pm

kellychoffman:
👍  Love it.

PR #169, opened 2016-05-26
New icon: Shipping classes

jameskoster:
shipping classes

Not entirely sure about this one. Would love some feedback.

"Shipping classes can be used to group products of similar type and can be used by some Shipping Methods (such as Flat Rate Shipping) to provide different rates to different classes of product."

The icon simply represents different types of products that might need different shipping rates.

¯_(ツ)_/¯

### DIFF
--- a/svg/gridicons-shipping-classes.svg
+++ b/svg/gridicons-shipping-classes.svg
@@ -1,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Generator: Adobe Illustrator 19.1.0, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->
+<svg version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px"
+	 viewBox="0 0 24 24" enable-background="new 0 0 24 24" xml:space="preserve">
+<g id="Adv._Guides">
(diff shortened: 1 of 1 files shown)


folletto:
Seems good to me, matched fairly well the meaning. 👍

drw158:
I like the idea, but again this won't work because Gridicons strictly follows a flat, two dimensional style. I think once we nail the "package" icon and its variants we can get this one too more easily.

Maybe something like this, but with varying package sizes: 
screen shot 2016-05-26 at 10 01 03 am

folletto:
> I like the idea, but again this won't work because Gridicons strictly follows a flat, two dimensional style. I think once we nail the "package" icon and its variants we can get this one too more easily.

Can we review this in the general "box" icon #168, since they all inherit the shape from there?

drw158:
@folletto Yes, I'm reviewing all of the PRs.

drw158:
Maybe we could use the folders icon for this

screen shot 2016-05-31 at 5 04 30 pm

jameskoster:
Whilst it's a difficult concept to inconise, I think the folders icon is a little too arbitrary imo. Could we apply the original idea (little & large product) to the new product icon?

drw158:
Ok, I'll give that a go.

jameskoster:
We may not be using this icon now so this can be put on ice.

PR #170, opened 2016-05-26
New icon: Shipping

jameskoster:
shipping

Based on the cart icon

### DIFF
--- a/svg/gridicons-shipping.svg
+++ b/svg/gridicons-shipping.svg
@@ -1,17 +1,1 @@
-<?xml version="1.0" encoding="utf-8"?>
-<!-- Generator: Adobe Illustrator 19.1.0, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->
-<svg version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px"
-	 viewBox="0 0 24 24" enable-background="new 0 0 24 24" xml:space="preserve">
-<g id="Adv._Guides">
-</g>
-<g id="Guides">
-</g>
-<g id="Artwork">
-	<g>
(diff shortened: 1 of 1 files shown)


folletto:
I think this needs some work, feels very odd. :)

Some random inspiration from a quick scan on google images:

screen shot 2016-05-26 at 14 44 29


screen shot 2016-05-26 at 14 44 38

kellychoffman:
Noun project has some [good trucks](https://thenounproject.com/search/?q=shipping%20truck) for inspiration. 

A box truck type would be appropriate for shipping I think. Something like the one below, but facing forward?

screen shot 2016-05-26 at 8 04 33 am

drw158:
Yes, I think a truck would be great for this. Your icon isn't quite there yet. Here is the work I've done for a truck (I'm not satisfied with it yet)

screen shot 2016-05-26 at 9 51 57 am

Material Design has a nice one:

screen shot 2016-05-26 at 9 52 43 am

folletto:
Your second one seems a good direction to me 👍

jameskoster:
> Your second one seems a good direction to me

I like that one too.

drw158:
Here is my new truck icon:

screen shot 2016-06-01 at 4 25 58 pm
screen shot 2016-06-01 at 4 25 35 pm

thoughts?

jameskoster:
I like it. Any reason you ditched the cut-out on the container?

folletto:
The shape is nice. I was wondering about the cut-out too, mostly because feels a bit too heavy. Here's a quick mock:

screen shot 2016-06-02 at 10 51 15

I guess a concern are the lower part with the wheels, where the line weight is more than our patterns standard. However it gives the van a more "van-like" look as well (the bottom part darker) so I'm not sure if we're beyond the limits here or not. :)

jameskoster:
I prefer that one personally.

drw158:
Thanks @folletto, I much prefer your icon. Updated PR:

screen shot 2016-06-02 at 10 32 34 am
screen shot 2016-06-02 at 10 32 19 am

PR #171, opened 2016-05-26
New icon: Tax

jameskoster:
tax


folletto:
Can we make these a little shorter? Tall like that seems a little too much skyscrapers to me. But again might be just me. :D

kellychoffman:
Yeah, having less of them would make it not so busy looking and easier to decipher at smaller sizes. It would also align better with the other icons in the group which are very simplistic, for lack of a better word.

drw158:
Unfortunately, this won't work because Gridicons strictly follows a flat, two dimensional style.

This one is tough; I would prefer something like the `%` or `$` symbol. In fact, your icon also represents "money" to me, so maybe we could come up with an icon for "price" or "money" and use it for tax.

Here's the open issue for price: https://github.com/Automattic/gridicons/issues/47

folletto:
Good point on depth. 👍

drw158:
Also, first thing I thought of was this balance thing, but I don't know if that represents tax or law.

screen shot 2016-05-26 at 9 54 37 am

jameskoster:
I want to stay away from specific currency symbols as a store can be using any currency.

I also thought about the % symbol. Wasn't sure the metaphor was too loose though. I guess a stack of coins isn't particularly great though. Will mock up a %.

jameskoster:
percent

drw158:
I think that could work. I'll do more research into the balance metaphor and see if there is anything we can use.

drw158:
Maybe we could use the future price icon for this

https://github.com/Automattic/gridicons/issues/47#issuecomment-222241841

jameskoster:
Definitely could, yes. Feel free to close.

PR #173, opened 2016-06-01
New icon: Price

### LINKED ISSUE #47: Add icon: Price

drw158:
screen shot 2016-06-01 at 12 15 11 pm

fixes #47 

cc @folletto


folletto:
Before merging I wanted to give it a go on replacing the middle circle with a face. What do you think?

comparison-with-face

If you feel it's a better one, let's go with the face. 
Otherwise, +1 for me to merge this one as-is. :)

drw158:
I think it _could_ work, but I am leaning towards the left one because we typically don't have that level of detail in the icons that the right one has.

Issue #155, opened 2016-05-19
Add icon: lightning icon


drw158:
For `offline`, and possibly `upgrade` in the future. And because it would be cool.

PR #179, opened 2016-07-05
New icon: Bolt

joanrho:
Here's a proposed lightning bolt for Issue #155 
bolt

In context with a few other Gridicons: 
bolt-in-context

### DIFF
--- a/svg/gridicons-bolt.svg
+++ b/svg/gridicons-bolt.svg
@@ -1,0 +1,1 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24"><title>gridicons-bolt</title><g id="Artwork"><polygon points="10.36 3 18 3 13.64 9 18 9 6 21 10.33 12 6 12 10.36 3"/></g></svg>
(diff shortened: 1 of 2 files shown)


mapk:
I did a quick review of how the angles are lining up with the other icons and it looks good! The weight of the icon feels great too.

icon-angles

joanrho:
Thanks @mapk!

Just added a resize icon, too:
resize

In context with other Gridicons:
resize-in-context

MichaelArestad:
This looks pretty dang good so far. It seems like a good candidate for the Upgrades icon, but probably not the power icon. The power icons generally tend to be pointy on both ends. Here's what we're currently using:

image

I also checked your icon at 1x. It looks pretty good, but might be a touch sharper if you can somehow align the points to whole pixels. (maybe not)

image

The resize looks pretty perfect, but I would remove it from this PR and add it in a second so we can discuss and merge one at a time.

joanrho:
Thanks for the feedback, @MichaelArestad! Good catch—the upper left "zig" corner and inner left and right "zag" corners seem to be slightly off whole pixels. Will update that!

bolt-grids

drw158:
A bolt icon was added in #183 after this PR was opened. 

We should keep this PR open, but know that it will replace the current bolt icon:

drw158:
will be handled in #187

PR #183, opened 2016-10-07
New icon: Lightning bolt.

jasmussen:
We have an offline lightningbolt icon, but it isn't a gridicon. It's just pretending to be. Let's make it official.

The old icon, by the way, had some slightly lazy pixels:

oldpixels

Here are the new pixels and newfound symmetry:

newpixels

In context:

screen shot 2016-10-07 at 16 00 17

CC: @rickybanister


jasmussen:
CC: @mtias you mention some generator tweaks needed in https://github.com/Automattic/wp-calypso/pull/8579 — I can do those as part of this PR too. What specifically needs to be done?

folletto:
Review:

* Style — looks good. It's different from almost any other shape, so it's rather unique, but I think the symmetry and straightening helps there. 👍 
* Semantic — I'm not sure we should add icons purely graphical. The lightning bolt was born to mean "offline" so I'd rather add it with the "offline" label, to mark its use and how it might change in the future if we need. But I'm also aware we might not have taken a clear stance if we want symbol-labelling or meaning-labelling here. We might want to take the decision and update the README. As a lightning bolt, it works. 

Gridicons checks:

* ✅  24dp base grid
* ⏺  straight 45 degree angles → _partially applicable given the subject_
* ✅  flat, bidimensional look (no faux 3D whatsoever)
* ⏺  2dp lines → _not applicable_
* ⏺  2dp radius rounded corners → _not applicable_
* ✅  icons should be sized optically so they are balanced against each other

We should discuss the labelling above, but apart from that it's good to go for me. :)

jasmussen:
I can rename it to offline if you like? I've no strong opinion.

folletto:
I'm ok either way, I just think we need to decide if we standardize on "symbolic" names (lightning) or "usage" names (offline), and do that for all the icons from now on then. :)

drw158:
My opinion on naming: unless there is absolutely no other use case for this icon, then it should be called "bolt". For example, the book icon is named generically because there are several uses for it. The bolt is a bit harder to pin down — there aren't a ton of uses, but I could see it being used for other things.  "Follow" obviously has no other use case, so it is named by usage.

I'm ok with the naming styles being different for each icon, and I don't see a benefit in having a consistent naming style.

jasmussen:
Hah, JUST renamed it to "offline" — easy to revert though. Are you set on it being "bolt", @drw158 ?

drw158:
I'm not set either way, I say just use your best judgement. Thanks Joen!

jasmussen:
Thanks Dave.

> unless there is absolutely no other use case for this icon, then it should be called "bolt"

By this logic, and disregarding any standardization rules for the time being, I think it should be called "offline". I would think using this icon for offline means it shouldn't be used for anything else.

folletto:
The argument in favour of _symbolic_ naming is that it can be reused multiple times, as it will also happen by chance.

The argument in favour of _usage_ is that it will allow us to update the icon in the future if we need to, without risking to break anything else because it's surely going to be used only for one thing.

Gruntfile.js
```diff
@@ -179,8 +179,7 @@ module.exports = function(grunt) {
 					"	},\n\n" +
 					"	render: function() {\n" +
 					'		return (\n' +
-					'			<div className="design-assets__group">\n' +
-					'				<h2><a href="/devdocs/design/gridicons">Gridicons</a></h2>\n';
+					'			<div>\n';
```
jasmussen: This change is being made to accommodate this bit of feedback: https://github.com/Automattic/wp-calypso/pull/8579#discussion_r82382104

rickybanister:
I support the 'bolt' name—we may want to use it in more than one situations (where it has more or less the same meaning) like you've lost your internet connection, but it could also be useful to signify that a remote Jetpack site is unreachable, thus 'offline'. Also, naming it 'bolt' makes it easier to find since that's what it looks like. I get the reasoning, but bolt seems better.

PR #184, opened 2016-11-17
Add docs page for future gh-page.

jasmussen:
This PR adds a new folder, `docs/`, which GitHub can reference as a GitHub Page to live at http://automattic.github.io/gridicons/. The benefit of doing it this way, as opposed to the current way where our GitHub Page lives in a `gh -pages` branch, is that we don't have to manually copy/paste new icons into that branch, as the `docs/` page is now automatically output by the grunt script we'll be running regardless.

Incidentally I also did a little cleanup on the page itself; removed the need for jQuery, and put both JS and CSS inline (since there's not a lot).


drw158:
Oh nice, will the gridicons.com domain still work?

jasmussen:
Yep, that just redirects. We could map it though, like we've done with genericons.com.

PR #189, opened 2017-01-31
Add <title> to minified SVGs 

### LINKED ISSUE #186: Accessability: Add <title> to minified SVGs
SVG allows us to add accessibility features to help screenreaders grok what an icon represents:

https://css-tricks.com/accessible-svgs/#article-header-id-6

We should add this. We already have the title of the icon, for example `gridicons-align-center`. We should add a section to the gruntfile that processes this into something like `<title>Align Center</title>`. Should be doable with a little regexing, space replacin' and titlecasin'!

drw158:
fixes #186 

This adds a <title> for accessibility, using the filename as the content.

I tried my hand at selecting words from the filename and title casing, but it's a bit out of my league.

Edit: Got some help and everything works.

- added a new grunt task, `addtitle`
- ran `grunt` to update all svg-min files

All svgs now have `title`

cc @jasmussen


jasmussen:
Thanks for working on this! 

I can probably help you out with the title casing and string polishing, have been a bit busy though. Will see if I can get to it next week!

Gruntfile.js
```diff
@@ -334,9 +334,13 @@ module.exports = function(grunt) {
 			// Grab the relevant bits from the file contents
 			var fileContent = grunt.file.read( 'svg-min/' + svgFile );
 
+			// Grab the filename
+			var name = svgFile.split( '.' );
+			name = name[0];
 
 			// Add transparent rectangle to each file
 			fileContent = fileContent.slice( 0, fileContent.indexOf('viewBox="0 0 24 24">') + 20 ) +
+						'<title>' + name + '</title>' +
```
gwwar: Maybe add this step to `svgmin` instead. `addsquare` is run last and appears to be used for sketch files?

Also there's lots of ways to do it but how about:
```javascript
var title = name.split( '-' ).splice( 1 ).map( function( item ) { 
    return item.charAt(0).toUpperCase() + item.slice(1); 
} ).join( ' ' );
```

gwwar: So for example if the name is `gridicons-add-image`, with the above should return `Add Image`

gwwar: Do we care about i18n here?

gwwar:
I would recommend rebasing with master, since I've modified the gruntfile this morning.

PR #190, opened 2017-02-01
Consolidate return value into common logic

aduth:
Related: #188 (Raised at https://github.com/Automattic/gridicons/pull/188#issuecomment-276535635)
Conflicts with: #189

This pull request seeks to shrink the size of `build/index.js` by stripping repeated SVG wrappers for individual icons and instead deferring the wrapper to the end of the function at the point where the component is returned.

Before|After
---|---
raw: 81.7kb|raw: 45.3kb
gzip: 16.3kb|gzip: 15.6kb

gzip sizes measured with [`gzip-size-cli`](https://www.npmjs.com/package/gzip-size-cli)

__WIP:__ This may or may not be worth pursuing, given disappointing results in GZIP size difference. I noticed there are some small visual regressions that need to be fixed before moving moving forward.

While the difference in size prior to gzip is huge (almost cut in half), based on how gzip works, it's not too unexpected that the repeated code had been heavily optimized. See also some related discussion at https://github.com/Automattic/wp-calypso/pull/8393#issuecomment-272034863.

Should we bother with this optimization?


folletto:
Good question, thanks for the PR too!

Reviewing the code, my feedback would be:

1. I find clearer to have a single instance at runtime where runtime values are set, instead of having the repetition done at build time. Not just more readable, but also it's easier to follow the logic.
2. Why is there a conditional logic to split `svg` and `d` into two different flows? Can't be just one making it more readable? I'm probably not understanding the need of a further optimization for `d`.

aduth:
>Why is there a conditional logic to split `svg` and `d` into two different flows?

Splitting brings the size down from 58.7kb to 45.3kb (15.9kb to 15.6kb gzipped, [diff](https://github.com/Automattic/gridicons/files/745415/remove-d.txt)).

To be honest, I don't really think we should care whether the generated build file is readable, so I'd favored size reduction over anything else.

folletto:
> To be honest, I don't really think we should care whether the generated build file is readable, so I'd favored size reduction over anything else.

Understood. :)

My opinion is that readability would be my only reason if we want to pull this in regardless of the `gzip` impact being null, so I'd change that for that reason. :D 

I agree: I don't see the benefit instead for pure size gain.

gwwar:
Looks like we're not getting a great win in size, as gzip works really well with lots of repeated characters. Perhaps we might want to focus on splitting out the icons into individual files?

aduth:
>Perhaps we might want to focus on splitting out the icons into individual files?

I agree the effort better spent 😄

PR #191, opened 2017-02-01
Build: add back svg-min, move react svg preprocessing to another folder

gwwar:
As requested this moves back the generated svg-min files into the repo, since folks were using these files directly.

## Testing Instructions
- `npm run build` 
- This should generate no local changes for git
- Adding an svg then running `npm run build` works normally
- temporary svg-min-react folder is cleaned up
- svgs in svg-min folder do not have transformed kebab case to camel case attributes.
Eg: `gridicons-user-add.svg` has `data-name` instead of `dataName`
- Generated react components do have camel case. eg search for `dataName` in `build/index.js`


gwwar:
Thanks for the review @folletto

aduth:
Not sure if you'd meant for this to be accompanied by a release, but current stable version on npm still does not have the `svg-min` folder.

folletto:
I don't think `npm` should include the individual icons. Is there a reason for that?

aduth:
Oh sorry, I must have misunderstood the intent of the PR. Disregard 😅

PR #194, opened 2017-02-13
New icon: Rewind

keoshi:
This adds a new Rewind icon.

We’ll eventually need an icon for the Activity Log Rewind action, so I just used the existing refresh icon and flipped it vertically. If it’s too ambiguous to have both, I’m happy to revisit and redesign something unique.

It’s not urgent that we merge this and, in fact, will not use it for a while. Just thought I’d give this a go while I’ll messing around with Gridicons. :)

Pixel view:

image

On page:

image


keoshi:
The more than I think about this, the more the history icon is looking like the right choice for this. I'll close this one for now, and reopen when we pick this up again.

PR #195, opened 2017-02-13
Add support for older IE browsers

keoshi:
Use SVG for Everybody library to support IE browsers up to version 9.

Removes Gridicons SVG inline code and replaces it with the external reference.

IE 9:

image


keoshi:
Yep, noticed that too, which could represent a little bump in testing but otherwise works fine. Thanks for reviewing!

PR #201, opened 2017-03-30
New Icon: Next Page

theck13:
Add a new icon for pagination representing the `<!--nextpage-->` tag.  This icon mimics the [gridicons-read-more](https://github.com/Automattic/gridicons/blob/master/svg/gridicons-read-more.svg) icon with the only difference being the dotted center line which copies the tag visualization in Calypso.  While the tag is supported in Calypso by entering `<!--nextpage-->` in the HTML tab, no icon currently exists.  This icon will be used by the WordPress Android and iOS apps in the format toolbar of the editor alongside the read more icon.

See the table below for comparison between the read more and next page icons.

Name | Icon | Calypso
-|-|-
Read More | ic_read_more | im_read_more
Next Page | ic_next_page | im_next_page

### DIFF
--- a/svg/gridicons-next-page.svg
+++ b/svg/gridicons-next-page.svg
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Generator: Adobe Illustrator 19.2.1, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->
+<svg version="1.1" id="Layer_1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px"
+	 viewBox="0 0 24 24" enable-background="new 0 0 24 24" xml:space="preserve">
+<path d="M3,5h18v4H3V5z M3,19h18v-4H3V19z M5,11H3v2h2V11z M9,11H7v2h2V11z M17,11h-2v2h2V11z M21,11h-2v2h2V11z M13,11h-2v2h2V11z"/>
+</svg>


jleandroperez:
@mattmiklic sir, may we bug you with this one?. Thanks in advance!!

mattmiklic:
> the dotted center line which copies the tag visualization in Calypso.

I like the logic behind the icon design. I'm not sure it'd be 100% obvious to me without clicking, though it would be obvious once you did, since it adds the "Page Break" text in the editor. This seems reasonable to me, but would be good if another of the designers with more experience creating Gridicons could weigh in.

drw158:
Whoa! I didn't realize we could do page breaks.

The difference between the dotted lines in Calypso doesn't seem significant, and looks a bit unintentional. Not sure about the history behind those images.

The icons, especially without a label, would be a bad case of mystery meat. For even experienced users, I don't think there is any way of inferring what those icons do. The Read More icon by itself is hard to decipher, even more so if there is another icon almost identical right next to it.

Here are some ideas. I think there is something to the idea of a literal break, or using the Page icon.

fullsizerender 9

mattmiklic:
Good suggestions @drw158! I'm kind of partial to this guy: 

screen shot 2017-03-31 at 1 36 48 pm

in that it re-uses the familiar page icon to give the icon a bit more context.

MichaelArestad:
I agree with @drw158 and dig some of those explorations. Particularly with the torn in half page.

folletto:
I agree the original icons aren't ideal in terms of semantic, and I think @drw158's exploration go in the right direction.

Maybe (this is an extra from the question above) I'd suggest actually to replace both more and page, and jumping of from the explorations above, I'm thinking something like:

moremore

theck13:
Both the [read more](https://github.com/Automattic/gridicons/blob/add/icon-next-page/svg/gridicons-read-more.svg) and [next page](https://github.com/Automattic/gridicons/blob/add/icon-next-page/svg/gridicons-next-page.svg) icons seem like an abstraction.  Looking at the progression below from the [next page](https://github.com/Automattic/gridicons/blob/add/icon-next-page/svg/gridicons-next-page.svg) icon on the left to the [pages](https://github.com/Automattic/gridicons/blob/add/icon-next-page/svg/gridicons-pages.svg) icon on the right makes sense, but the abstract icon on the left isn't intuitive to me by itself.  Reusing the familiar [pages](https://github.com/Automattic/gridicons/blob/add/icon-next-page/svg/gridicons-pages.svg) icon does help.



The solid line through the icon shown below seems like a thematic break or horizontal rule tag (i.e. `<hr/>`) which will also be supported in the WordPress apps and is currently represented in Calypso by something similar to the [minus small](https://github.com/Automattic/gridicons/blob/add/icon-next-page/svg/gridicons-minus-small.svg) icon.

hr

Would these three icons be easily recognizable for page break, read more, and horizontal rule?

nprmhr

Note:  As for the lack of labels and mystery meat, long-pressing on the format button reveals the action like "Read More" and "Page Break" for example.

folletto:
Feedback on that:

1. I wouldn't change "horizontal rule", as it's an element as others, and it thus doesn't require the "page" to be shown in the icon, like "image", "link" or "quote" doesn't show the page in the icon either.
2. I would totally use either of the three above for "more" as it literally "breaks" the page in two parts accessible by a click. They are spot on for this meaning 💯 
3. I would try to show two pages in the "page break", as I think it would be misleading to show just a single page that gets split in two, as in practice the result is two pages, not one split like more.

drw158:
I like the idea of these:

screen shot 2017-04-03 at 9 26 40 am

Left being `read more`, the right being `page break` or `next page` (?). 

I agree with folletto that the `hr` doesn't need the context of the page, and makes more sense to treat it like content. Although I do get the reasoning — it's a gray area because `hr`s do separate content similarly to `read more`. I don't feel to strongly about it. Perhaps the minus icon is unclear and could use the context of the page icon.

I think it'd be worth clarifying `page break` vs `next page`. When I think of `page break` I think of the post being truncated with a read more link. As in, "Read the rest of the post after the break". I actually think `page break` is much more clear that `read more`.

If splitting the page into several numbered pages is indeed a `next page`, then I think that would lead to a different icon design, probably something like folletto drew, with 2 pages.

maybe something like 

screen shot 2017-04-03 at 9 54 40 am

folletto:
> Perhaps the minus icon is unclear and could use the context of the page icon.

Yeah I agree the minus icon can be improved, I just don't think that grouping it with the rest is effective for a purely visual page element (that some themes might even visualize differently).

> maybe something like

+1 for me, that seems a clearer differentiation.

mattmiklic:
> Maybe (this is an extra from the question above) I'd suggest actually to replace both more and page, and jumping of from the explorations above, I'm thinking something like:

> maybe something like

I really like this direction; where `read more` uses the split page metaphor and `next page` shows multiple sheets, which mirrors the paging of the article.

diegoreymendez:
Any updates on the state of this icon?

It's soon going to be visible to a wider audience (through the upcoming Aztec closed and open betas), so it may be interesting to include something closer to the final icon if possible.

drw158:
Ppl seems to agree that the multi-page icon would represent "next page" best. How is this?

screen shot 2017-04-20 at 11 23 38 am

folletto:
Looks good. 👍 

Maybe tweak the lower line to be shorter, to match the top page (due to its folding)? But it's just a tweak — I would ship the above as is too.

jleandroperez:
Pinging @drw158 (OR) @folletto! gentleman, can we merge any of the proposals in this thread?

Thanks in advance!

mattmiklic:
Since this icon is being talked about in the context of the editor's formatting toolbar, let's loop in @iamthomasbishop since he's working on the design of that UI.

folletto:
As mentioned above — we can proceed with the last one. 

@drw158 you have the vectors, want to push them? Otherwise just share them and @iamthomasbishop or I can merge them :)

iamthomasbishop:
I'm a fan of this approach, atm: https://cloud.githubusercontent.com/assets/618551/24615332/93a9574e-1853-11e7-8bc4-f5553ea0ddcf.png

but I'd be curious to test with a sizable group of users (maybe this has already been done?) to see what the consensus is. I'm not sure what our usual approach is here.

folletto:
The approach I'd take there is: right now we don't have any, so the most important thing is to include them in Gridicon and ship them – I agree the two you have there are the best alternative right now.

That will give us the baseline, and allow us to test it. :)

folletto:
@drw158 should we open a separate one for "Read More", as I think your design above is an excellent improvement to the one we have now that is easy to confuse with horizontal line?

drw158:
I'll make one!

jleandroperez:
Thank you @drw158 + @folletto + @mattmiklic + @theck13!!!

:bowtie:

PR #203, opened 2017-04-14
Docs: specified ideal resolutions and exception for my-sites

folletto:
After a discussion (puPv3-4jW-p2) we noticed we should be more clear about:

1. the ideal sizes of the icon set
2. that `gridicons-my-sites` is a special case of the WordPress logo, and shouldn't be used larger than 36px.

This PR edits the readme with this note:

> **Notes**: the icon set is made to be used exactly at these pixel sizes: 12, 18, 24, 36, 48, 54, 72. However, `gridicon-my-sites` as it's a small-size version of the WordPress logo, shouldn't be used larger than 36px.

/cc @mattmiklic

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,10 @@ render() {
 * `size`: Number - (default: 24) set the size of the icon.
 * `onClick`: Function - (optional) if you need a click callback.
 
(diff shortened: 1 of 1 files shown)


folletto:
Following @jasmussen's advice we added a link to Social Logo's repository:

> **Notes**:
> 
> * The icon set is made to be used exactly at these pixel sizes: 12, 18, 24, 36, 48, 54, 72.
> * `gridicon-my-sites` as it's a small-size version of the WordPress logo, shouldn't be used larger than 36px. If you need to use the WordPress logo in larger sizes, see the [Social Logos project](https://github.com/Automattic/social-logos).

PR #204, opened 2017-05-09
Adopt React 15.5+ approach to PropTypes

stuwest:
React 15.5+ [moved PropTypes out of the main `react` package and into a separate `prop-types` package](https://facebook.github.io/react/blog/2017/04/07/react-v15.5.0.html#migrating-from-react.proptypes). I'm using `gridicons` for a project and it's now the last dependency that is still throwing off deprecation warnings.

This PR attempts to address. It builds without errors, and addresses the deprecation warnings in my project, but could use thorough testing by someone more familiar with this code and packaging up modules for npm (e.g. should `prop-types` be in `peerDependencies` or `dependencies` see [migration guide](https://github.com/reactjs/prop-types#migrating-from-reactproptypes)). We may have to bump React version to ^0.14.9.

A second commit cleans up some minor linting issues based on [the Calypso ESLint config](https://github.com/Automattic/eslint-config-wpcalypso).

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,6 +18,9 @@
     "build/index.js",
     "build/example.js"
   ],
+  "dependencies": {
+      "prop-types": "^15.5.10"
+  },
   "devDependencies": {
     "babel-plugin-add-module-exports": "^0.2.1",
     "babel-plugin-syntax-jsx": "^6.18.0",
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
     "prepublish": "npm run build"
   },
   "peerDependencies": {
-    "react": ">=14.x"
+    "react": "^15.3.0"
   },
   "repository": {
     "type": "git",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "gridicons",
-  "version": "1.1.0",
+  "version": "2.0.0",
   "main": "build/index.js",
   "scripts": {
     "build": "grunt --verbose",
@@ -19,7 +19,7 @@
     "build/example.js"
   ],
   "dependencies": {
-      "prop-types": "^15.5.10"
+      "prop-types": "^15.5.7"
   },
   "devDependencies": {
     "babel-plugin-add-module-exports": "^0.2.1",
--- a/sources/react/index-header.jsx
+++ b/sources/react/index-header.jsx
@@ -10,7 +10,8 @@ OR if you're looking to change now SVGs get output, you'll need to edit strings 
 /**
  * External dependencies
  */
-import React, { PureComponent, PropTypes } from 'react';
+import React, { PureComponent } from 'react';
+import PropTypes from 'prop-types';
 
 export default class Gridicon extends PureComponent {
 


folletto:
> A second commit cleans up some minor linting issues based on the Calypso ESLint config.

FWIW: I would personally split these as it makes a clearer PR if we address just one issue at time —
 but it's ok as it's already done here and it's fairly simple.

---

/cc @gwwar maybe can help reviewing?

package.json
```diff
@@ -18,6 +18,9 @@
     "build/index.js",
     "build/example.js"
   ],
+  "dependencies": {
+      "prop-types": "^15.5.10"
```
gwwar: Are we ready for this? Calypso isn't on React 15.5 yet, I would recommend moving this to a peer dependency, and bump the react peer dependency to >=15.5.

The package version should probably be bumped to `"version": "2.0.0-alpha.1"`

gwwar: One option might be to tidy up this branch, then publish an 2.0 alpha version in npm under the next tag. So you'd be able to install this newer version with `npm install gridicons@next`

gwwar:
I would recommend splitting out 09383e0 into it's own PR since there's nothing blocking that from merging into master.

stuwest:
> I would recommend splitting out 09383e0 into its own PR since there's nothing blocking that from merging into master.

https://github.com/Automattic/gridicons/pull/223

package.json
stuwest: > Are we ready for this? Calypso isn't on React 15.5 yet

This particular version number is just for the new `prop-types` component, which while it mirrors some React versioning is a separate thing. According to its [migration guide](https://github.com/facebook/prop-types#compatibility), `prop-types` is compatible with 15.3.0 and up, so [the 15.4.0 in Calypso](https://github.com/Automattic/wp-calypso/blob/master/package.json#L106) should be fine.

It looks like it works with either 0.14.9 and higher or 15.3 and higher. So it seems at a minimum we'd want to bump our `react` dependency to 14.9. Given Calypso is already 15.4.0, maybe we just push it to 15.3.0? Added that in a separate commit and rebased to master.

> I would recommend moving this to a peer dependency

I don't really understand the alternatives too well, but the `prop-types` [migration guide](https://github.com/facebook/prop-types#how-to-depend-on-this-package) suggests leaving it in `dependencies`.

gwwar: I'm trying to avoid any dependencies in this library, since this is a simple icon lib. 

I don't think it's too much of a burden for users of this lib to have React and PropTypes as dependencies. This is why when we update the package to use prop-types in our component, we'll need to bump the major version since the requirements have changed.

gwwar: @stuwest How about we update this to lowest compatible version `^15.5.7` so more folks have a chance of de-duping the dep?

We'll also still want to bump the main package version from `1.1.0` to `2.0.0-alpha.1`  since we're jumping a peerDep from 14.x to 15.x and open a Calypso PR so we can test the change. We can either publish an alpha package with `npm publish --tag next`, where on the Calypso branch we install with `npm install gridicons@next`. OR we can point at the git url directly in package.json and the shinkwrap. The link for this branch would be `Automattic/gridicons#fix\/prop-types`

Before we merge this, we can set the version back to `2.0.0` and do a publish after it lands in master.

gwwar:
I'll rebase and test this once #234 is in

sirreal:
> I'll rebase and test this once #234 is in

@gwwar ✅  😀

gwwar:
Here's the related Calypso PR https://github.com/Automattic/wp-calypso/pull/15758 help testing is appreciated.

gwwar:
Thanks @stuwest !

Issue #205, opened 2017-05-19
Cross gridicon: omit title element?


bluefuton:
I've noticed that when we render the cross gridicon in Calypso, the SVG `<title>` (in this case `gridicons-cross`) is shown as a tooltip in Chrome.

https://raw.githubusercontent.com/Automattic/gridicons/master/svg/gridicons-cross.svg

Should we omit the title element where it just reflects the icon name?

![2017-05-19 16_39_37](https://cloud.githubusercontent.com/assets/17325/26252985/17ed855c-3cb2-11e7-8296-e5431813c511.gif)

folletto:
I'm not even sure it makes sense to include a title now that you point it out. Afaik it's [used for accessibility reasons](https://developer.mozilla.org/en/docs/Web/SVG/Element/title), but this is an abstract library, and as such:

1. If used in a decorative way, they probably shouldn't show up as interactive
2. If used in an active way, they should have a title that is "noun+verb" of the action to be performed, not the iconography of the item

Either way, it seems something that the app using gridicons should be doing, not gridicons themselves. 

Unless I'm missing something, clearly.

drw158:
Seems like a bug. You can see that the title of the icons are formatted correctly in the svg-min folder.

https://github.com/Automattic/gridicons/blob/master/svg-min/gridicons-cross.svg?short_path=b5df1b0#L1

For example, title of this icon is `Cross` not `gridicons-cross`.

So either the Gridicons package was not updated in Calypso, or Calypso is injecting its own title?

mtias:
> We need to keep the title for accessibility.

This was raised for dashicons, and the argument seems to be that it's the opposite, titles should be removed from the svgs: https://github.com/WordPress/gutenberg/issues/528

The icon itself doesn't have a meaning, it's when it's wrapped as a button, etc, that it takes one, and then it's the button the one that should have an aria role.

folletto:
Ok let's split the two discussion for clarity :)

# Accessibility

Dave and I discussed a bit to clarify the positions above, and we both think some extra clarity is needed here: on one hand, `title` exists explicitly to be accessible, so it's good to have. On the other hand, "decorative images" and "images as part of a link" should have the alternative text void (see the [WAI tutorial here](https://www.w3.org/WAI/tutorials/images/decorative/)). 

I'd personally lean more in not including title, and @mtias link seems confirming it. I'd like to have  @jasmussen as he suggested it in #186 for cross-check.

# Minified or Not?

[...]

Issue #123, opened 2015-12-04
Build: use a DOM-aware library to manipulate SVGs


folletto:
Following from the discussion in #122, would be better if we were able to replace direct string manipulation with XML (SVG) aware libraries, such as [cheerio](https://github.com/cheeriojs/cheerio). This would make the code more reliable in the long term.

Example of direct string manipulation: 

```
    fileContent = fileContent.slice( 0, -6 ) +
        '<rect x="0" style="fill:none;" width="24" height="24"/>' +
        fileContent.slice( -6 );
```

PR #207, opened 2017-05-20
Fix: Remove <title> tags from SVGs

mladenplavsic:
As described in #205 <title> tags are removed in these usages of SVG:

- svg-min/*.svg
- php/gridicons.php
- svg-sprite/gridicons.svg

This is done by updating ```Gruntfile.js``` - removed "addtitle" task.

Some more things done:

- Updated ```grunt-svgstore``` to version ```1.0.0``` in ```package.json```
- Added option ```includeTitleElement``` and set to ```false``` https://github.com/fweinb/grunt-svgstore#optionsincludetitleelement-since-050
- Removed unnecessary option ```prefix```
- Using ```svg-min/*.svg``` instead of ```svg/*.svg``` for "svgstore" task


folletto:
`includeTitleElement = false`? Clever. :)

I'm a bit uncomfortable with `fileContent.indexOf( '>' )` which seems too generic, but as we use GIT/PRs if there are odd results should be visible. For reference, I already raised the point in the past, so it's just to make sure we're aware, and there's an issue to fix it: #123. 

---

To test, I checked:

1. No more `title` in svg-min/*.svg ✅ 
2. No more `title` in php/gridicons.php ✅ 
3. No more `title` in svg-sprite/gridicons.svg ✅ 
4. Run `grunt`, no differences expected ✅ 

_Note:_ I had to run `npm install` before running `grunt` for this to work properly.

folletto:
@bluefuton can you drop a quick eye here on the code to have a more expert review? Should be quick :)

drw158:
My PR #208 caused the conflicts. Looks like this collapses the sprite into one line. Note that I added a new icon in the PR.

bluefuton:
Thanks @mladenplavsic! Code looks good to me - just needs rebasing and SVGs regenerating I think 👍

drw158:
@mladenplavsic feel free to squash and merge.

mladenplavsic:
@drw158 I would like to, but I don't have permission for it :-)

bluefuton:
I'll do the honours. Thanks @mladenplavsic!

PR #209, opened 2017-05-24
Update icon: read more

drw158:
This updates the Read More icon to be more descriptive.

screen shot 2017-05-24 at 3 07 03 pm

screen shot 2017-05-24 at 3 02 30 pm

Previous:

screen shot 2017-05-24 at 3 06 13 pm


drw158:
Couple of options:

screen shot 2017-05-24 at 3 10 01 pm

Far right is the one is this branch.

folletto:
I'm ok with all the three icons. I'd probably have a slight preference for the first one as it's cleaner, so I'd suggest to consider it, but either way I think it's good. 👍

drw158:
Ok, added the svg.

screen shot 2017-06-09 at 3 57 35 pm

folletto:
Looks great! 👍 

When you merge, just be careful that the current SVG contains: `<title>1Artboard 1</title><g id="Artwork">`, let's double check it gets generated properly. 
I think this is another thing the new updated process we're doing in #217 will make easier. :)

PR #213, opened 2017-05-26
Remove babel transform-runtime plugin

### LINKED ISSUE #212: Add babel-runtime to dependencies
I tried to create a new nodejs Electron app with Gridicons by adding the package using npm. When run, I get this error: `Cannot find module 'babel-runtime/core-js/object/get-prototype-of'`. 

It is, indeed, at the very beginning of the compiled Gridicons `build/index.js` file after building:

```
var _getPrototypeOf=require('babel-runtime/core-js/object/get-prototype-of')
```

Installing the `babel-runtime` package from npm solves the issue.

I'm not sure why it's required, but perhaps it should be a dependency or a peer dependency?

gwwar:
Fixes #212

I had copy pasta'd the Calypso config, and missed that https://babeljs.io/docs/plugins/transform-runtime/ requires a hard dependency of `babel-runtime`. I went ahead and removed it since I'm not sure if we need these polyfills

For testing convenience, I published this to `1.1.0-alpha.1`. So you can install this via: `npm install gridicons@next --save`

cc @aduth @sirbrillig

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -209,7 +209,6 @@ module.exports = function( grunt ) {
 				],
 				comments: false,
 				plugins: [
-					'transform-runtime',
 					'transform-class-properties',
 					'transform-export-extensions',
 					'add-module-exports',
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "gridicons",
-  "version": "1.0.1-alpha.1",
+  "version": "1.1.0-alpha.1",
   "main": "build/index.js",
   "scripts": {
     "build": "grunt --verbose",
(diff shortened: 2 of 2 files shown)


aduth:
Do we need the runtime polyfills?

https://www.diffchecker.com/TsUjTIyo

I'm thinking... probably not. If anything, ES5+ on `Object.create` and `Object.getPrototypeOf`.

gwwar:
See also https://github.com/Automattic/wp-calypso/pull/14531

gwwar:
Alright, I tested a bit and I think we'll be okay with this change. I'm going to go ahead and merge this, then publish 1.1.0

gwwar:
@sirbrillig 1.1.0 is now available, so we can install via `npm install gridicons@latest --save` or `npm install gridicons@1.1.0 --save`

Issue #206, opened 2017-05-19
Run svgmin on the svg folder in addition to the svg-min folder


drw158:
The svgs in the svg folder vary a lot in content and formatting, and ideally they should be pretty simple, like

```
<svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24">
    <path d="M19 13h-6v6h-2v-6H5v-2h6V5h2v6h6v2z"/>
</svg>
```

As raised in #205 we don't need the title element.

This causes a change in our workflow. Currently , we say to export the svg to the svg folder in AI by using the export script. 

https://github.com/Automattic/gridicons/blob/master/README.md#add-a-proposed-icon-to-gridicons-admins-only

Perhaps we should also run svg min on the svg folder? Perhaps we could just get rid of the svg folder, and only have the svg-min folder?

I raise this issue because I think the sprite script uses the svg folder instead of svg-min folder. It doesn't seem wise to keep around a folder of messy svg files.

folletto:
> Perhaps we could just get rid of the svg folder, and only have the svg-min folder?

I'd probably go for this option, as I agree we shouldn't keep around a folder we don't use.

My only question would be if we have any use of the non-compressed folder... as it's just an output like `svg-min`, the sources are elsewhere, and `svg-min` seems preserving the invisible layers and shapes in graphic editing software... I think we can safely remove the `svg` only one.

drw158:
Hm, perhaps `svg` could be the exact same as `svg-min` except with the invisible rect added. We don't need the invisible rect in `svg-min` since it's just for graphic editing software. Then, one folder would be truly minified, and the `svg` could be used for mockups.

folletto:
This should be solved by the change in #217 right? Can we close?

drw158:
Illustrator is now out of the process, but I still propose we:

- minimize the `svg` folder with grunt
- add the invisible rect to the svgs in the `svg` folder
- remove rect from the `svg-min` folder

The svgs in the svg folder are unnecessarily messy (AI comments, extra code, title, etc.)

I'll update the title of this PR.

folletto:
Ok for me... even if I'd have opened a new PR as it would have better preserved history and would have been easier to read. But it's ok ;) 

> Or perhaps, we don't even need 2 svg folders?

Yeah, I agree with this bit: the `svg` folder is now a source folder, so we shouldn't touch it with automated scripts.

But as you said...

[...]

Issue #23, opened 2015-03-30
Ai doesn't allow for more than 100 artboards


davewhitley:
We'll need to adjust gridicons.ai so that it doesn't need 1 artboard per icon.

Issue #19, opened 2015-02-27
Move icon development from Illustrator to Sketch?


jasmussen:
Right now all gridicons are designed in Illustrator. Each icon has an artboard on its own. Each is saved as SVG and exported, then renamed and grunted.

The process is not super difficult, but it's also not super smooth either. Would migrating to Sketch improve the situation here? 

I'm somewhat new to Sketch myself, but it seems like Sketch would allow for somewhat simpler exporter options, not only to SVG files for the grunt process, but even for exporting multiple densities of each icon for Android and iOS at the same time. Seems like there's a benefit there.

Both have the pain point that adding in the icons is sort of a manual merge. 

Thoughts?

PR #217, opened 2017-06-09
Refactoring: simplified design/build process, SVG as master

folletto:
screen shot 2017-06-09 at 18 16 56

This PR is a proposals discussed with @drw158, @jasmussen, and @mattmiklic to simplify the process to add and update Gridicons.

Changes:

1. The master files are SVGs inside `/sources/svg/`
2. **No more AI files** — except for the template
3. **No more Illustrator scripts**
4. Since I was there, I also reviewed the comments in the `Gruntfile.js` to make it more readable.
5. Should also address the question raised in #206: instead of removing `svg/` we remove the AI files, the end result is more tight source SVGs.

Note that it seems a huge PR but most of the 180+ files are just files moved around.

### Benefits

* Faster turnaround of PRs as icons can be more easily merged
* Easier for anyone to add new icons
* Less binary files to handle
* Complete automation is possible

The only slight "con" is that SVGs might lose some editing capabilities the IA have, but since we started Gridicons in 2014 this issue never appeared in practice, and if anything we have used SVG as design sources.

### Historical Note

Initially we wanted to generate everything from Illustrator using Artboards, but we hit Illustrator limit to 100 Artboards (yes, really) and we reverted to single stack in #23. 
We also tried to use Sketch to preserve the ability to see all the icons side by side, but we had issue with that too, see #19 (mostly due to Sketch being not too evolved in SVG export at the time, today might be different).

Android also moved to use SVGs directly (with Android Studio) and iOS still requires a manual conversion to PDF, none of which again requires an Illustrator file. Due to these updated processes, the need for the special `gridicons-32.ai` file is reduced as tweaks should be done on a per-icon basis (see #14), and even there, we can just make `32` versions of the SVG files.

Thus, as long as the template is present to make sure the margins are followed (was added by Dave in #24), our real "master" copy of the icons are the SVG. 

### To test

1. Delete all temp folders: `node_modules`, `build` (even better: a clean git repository)
2. Readme: follow the "Propose a New Icon" steps (without creating a PR)
3. Readme: follow the "Add a Proposed Icon to Gridicons (Admins Only)" steps (without merging any PR)
4. Verify that all the necessary files are re-built as before (`svg-min`, `build`, `php`, `svg-sprite`, `docs`)
5. Try to remove the added icon (from both `sources/svg` and `svg-min`) and check in the diff that everything is restored to the initial state.


drw158:
I only got tripped up trying to find the `svg` folder. I just pushed an update to the readme to make that more clear.

Here is what was generated with `grunt`

screen shot 2017-06-09 at 3 35 05 pm

Removing the icon from svg and svg-min and grunting, cleanly reverted back to the initial state.

I don't think anything gets generated in `react`? I don't see anything in that folder that needs to be updated. Not sure.

drw158:
Also, we refer to the gridicons-32 file in the readme. Do we still need it? Were the icons in there any different than the 24px icons? I think they were slightly tweaked for mobile.

folletto:
> I only got tripped up trying to find the `svg` folder. I just pushed an update to the readme to make that more clear.

Thank you, I missed that.

> Everything looked good.
> Removing the icon from svg and svg-min and then grunting, cleanly reverted back to the initial state.

Yey! :D 

> I don't think anything gets generated in `react`? I don't see anything in that folder that needs to be updated. Not sure. We might need to update the readme to reflect the new stuff that gets generated.

Good point. I'll look into that.

> Also, we refer to the gridicons-32 file in the readme. Do we still need it? Were the icons in there any different than the 24px icons? I think they were slightly tweaked for mobile.

They were slightly tweaked for sharpness, as mentioned I'm not sure if we need them anymore, but for good measure, I added them back as SVGs in `sources/svg-32`.

Double checked them for sharpness:

screen shot 2017-06-10 at 21 56 46

folletto:
> I don't think anything gets generated in `react`? I don't see anything in that folder that needs to be updated. Not sure. We might need to update the readme to reflect the new stuff that gets generated.

Yes. The script has been modified. `react` contains a source template, the output is in `build`. Was changed in #188. For the scope of this PR I won't change it, I'll do a separate one to clear that up too. I need some extra clarity on the build / publish to npm to do that too. I'll follow up later.

For the time being, I updated the `README.md` file with these details.

folletto:
For clarity, I moved the React templates from `react` to `sources/react`. I checked regenerating the `build` and verified all the files with `md5` and they all match 1:1.

As a reference, currently the npm `package.json` uses only:

```
...
  "files": [
    "build/index.js",
    "build/example.js"
  ],
...
```

folletto:
Ok, all the feedback should be now addressed:

1. React folders clarified ✅ 
2. Gridicons for 32px added as SVG sources ✅ 
3. Output and notes clarified in the readme ✅  

Ready for another review. 👀

drw158:
Everything seems to work. Here are the files that get generated with grunt (excluding the initial svg)

screen shot 2017-06-12 at 9 49 37 am

mattmiklic:
> They were slightly tweaked for sharpness, as mentioned I'm not sure if we need them anymore, but for good measure, I added them back as SVGs in sources/svg-32.

Yeah, the iOS navbar requires 32px icons, and when we tried scaling up from the originals, they weren't quite sharp enough. We probably don't really need the `gridicons-32.ai` file, since it's only those five icons that need the size adjustments, and if we ever end up changing one of the navbar icons it, too, would be a one-time adjustment.

folletto:
Ok, so I think preserving for now the SVG version is good for now. :)

Double thanks @drw158 and @mattmiklic! 💯 

Merging the beast so we can start enjoying the benefits. :)

PR #219, opened 2017-06-09
New icon: share computer (social preview)

### LINKED ISSUE #181: New icon: Reader Card
We'd love to add this icon for previewing what a post will look like in the reader, and other services that offer a card view.

What do we need to do to make this fit the style guidelines?

19d4e410-639f-11e6-9351-62ecd33fb48a

drw158:
Note the name `share-computer`

I thought it would make sense to group it with the share icons. `social-preview` feels a bit to granular, and I think it should be able to work generically.

screen shot 2017-06-09 at 4 12 00 pm
screen shot 2017-06-09 at 4 16 33 pm

screen shot 2017-06-09 at 4 16 53 pm

fixes #181


folletto:
The icon looks good. 
I also tried to tweak the inner "share icon" centering just to be sure, and the one you choose I think it works best. :shipit: 

I also agree on the name suggestion, both for making it grouped together and more general clarity 👍

PR #225, opened 2017-06-16
PropTypes package

adambbecker:
Got a console warning when using Gridicons as a package that reads:

> Warning: Accessing PropTypes via the main React package is deprecated, and will be removed in  React v16.0. Use the latest available v15.* prop-types package from npm instead. For info on usage, compatibility, migration and more, see https://fb.me/prop-types-docs

This PR addresses that and moves PropTypes out of the React package and requires it as a "peerDependency".

*Note:* Not entirely sure if this & React should be `peerDependencies` vs. normal `dependencies`, but placed PropTypes there because React was already listed in that category.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,8 @@
     "prepublish": "npm run build"
   },
   "peerDependencies": {
-    "react": ">=14.x"
+    "react": ">=14.x",
+    "prop-types": "^15.5.10"
   },
   "repository": {
     "type": "git",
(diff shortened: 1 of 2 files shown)


gwwar:
Hi @adambbecker funnily enough we have #204 already open, but we could use some help on the Calypso testing side if you're interested

adambbecker:
@gwwar Oops, sorry about that, just saw that as well. Closing this one.

PR #226, opened 2017-06-16
Build: React, add "... otherProps"

adambbecker:
Adds support for other React props to be passed to the icons such as title, style, etc.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -317,7 +317,7 @@ module.exports = function( grunt ) {
 
 			// Add className, height, and width to the svg element
 			fileContent = fileContent.slice( 0, 4 ) +
-						' className={ iconClass } height={ size } width={ size } onClick={ onClick }' +
+						' className={ iconClass } height={ size } width={ size } onClick={ onClick } { ...otherProps }' +
 						fileContent.slice( 4, -6 ) +
 						fileContent.slice( -6 );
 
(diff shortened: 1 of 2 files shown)


gwwar:
Looks like we have a few other changes we'll like to queue up for another release. I'm going to go ahead and merge this @adambbecker

Gruntfile.js
```diff
@@ -317,7 +317,7 @@ module.exports = function( grunt ) {
 
 			// Add className, height, and width to the svg element
 			fileContent = fileContent.slice( 0, 4 ) +
-						' className={ iconClass } height={ size } width={ size } onClick={ onClick }' +
+						' className={ iconClass } height={ size } width={ size } onClick={ onClick } { ...otherProps }' +
```
gwwar: @adambbecker looks like we can override classname, so we'll want to add special handling for that prop

cc @folletto

sources/react/index-header.jsx
```diff
@@ -150,5 +150,5 @@ export default class Gridicon extends PureComponent {
 
 		switch ( icon ) {
 			default:
-				svg = <svg height={ size } width={ size } />;
+				svg = <svg height={ size } width={ size } { ...otherProps } />;
```
gwwar: Ah reading though again we'll want to omit className since we do have special handling for it already

adambbecker:
@gwwar Ah, I see. Yeah it looks like #235 (thank you!) provides a fix for this. Is that the case or should we reopen this?

gwwar:
Yes, @adambbecker #235 should should fix this. 😄

PR #227, opened 2017-06-19
Build: Update example site with display controls

drw158:
This adds some display options to the example site.

I just remembered that these changes need to be moved to the build script.

screen shot 2017-06-19 at 12 31 27 pm


folletto:
I think the changes need to be added here: https://github.com/Automattic/gridicons/blob/master/Gruntfile.js#L57

I plan to refactor that as an external template in a PR I've upcoming, so it's clearer where the code is coming from instead of being mashed inside the `Gruntfile.js`.

drw158:
Ok! I've moved the changes to the gruntfile, but when I run grunt, for some reason all of the svgs in svg-min get regenerated. Looking into it now.

folletto:
I tried to rebuild everything, looks good. No diffs. 
The page also loads as expected.

🚢

PR #231, opened 2017-06-20
Update Readme file to point to Gridicons example

SergioEstevao:
Update readme to point to specific Gridicons sample gallery.


drw158:
ok 👍

PR #234, opened 2017-06-27
User add: Remove data-name and clean up

sirreal:
All changes relative to the `user-add` icon. Does not affect display although SVG structure is modified.

* Removes `data-name` attribute (causing warnings from React).
* Removes extraneous `<g>` wrappers.
* Fuses some paths.

The most relevant change is ba87a54 which should fix React warnings, the rest is cleanup.

This was prompted by an error when using the `user-add` Gridicon within a React context (Calypso).

The following warning is generated:

```
Warning: Unknown prop `dataName` on <g> tag. Remove this prop from the element. [...]
```

warning

**Master:**

master

**Branch:**

change


drw158:
This is probably my fault when I exported the icon. I think the issue is either

1. For some reason, the build script is using the svg folder instead of the svg-min folder. 
2. The minify script is not stripping out unnecessary g elements and data name attributes.

sirreal:
> The minify script is not stripping out unnecessary g elements and data name attributes.

@drw158 I noticed this. [This source](https://github.com/Automattic/gridicons/blob/1cc4c1d9f25ec84d2d7d3ca9ecc88d10866b96a0/sources/svg/gridicons-user-add.svg?short_path=f74ed90#L2-L4) actually has [a `<g>` added](https://github.com/Automattic/gridicons/blob/1cc4c1d9f25ec84d2d7d3ca9ecc88d10866b96a0/svg-min/gridicons-user-add.svg?short_path=a6e0ff8#L1) at build.

Beyond the scope of this PR, but I was curious about the status of this repo, it does seem like it could use some ❤️  around builds and probably elsewhere.

sirreal:
Anecdotally, source SVGs

gwwar:
> It does seem like it could use some ❤️ around builds and probably elsewhere.

@sirreal PRs are always welcome. I did the minimal amount during a hack day so we no longer have to copy 🍝  components manually into Calypso.

I can help test this tomorrow. We can probably try and get together another release around this and PropType fixes

folletto:
> This is probably my fault when I exported the icon. I think the issue is either

I think it's just that we relied on software exports before and we didn't "review" the SVG. Now we can review it, so these things pop up and can be easily fixed. :)

> For some reason, the build script is using the svg folder instead of the svg-min folder.

I double checked and it's using `svg-min` — however I might not be getting what you meant here.

Note: React requires a temporary "conversion" of `svg-min` to `svg-min-react` as its XML syntax is different (see [here](https://github.com/Automattic/gridicons/blob/master/Gruntfile.js#L412)).

> The minify script is not stripping out unnecessary g elements and data name attributes.

I think this is less a minify issue and more how we have the original SVG files, as there might be instances when we want `<g>` to stay.

> This source actually has a <g> added at build.

This is a different matter, that `<g>` afaik is added to be able to target the entire icon (either code, or inside editing software). We can review that decision tho. The task `group` in Grunt does it ([ref](https://github.com/Automattic/gridicons/blob/master/Gruntfile.js#L387)).

> It does seem like it could use some ❤️ around builds and probably elsewhere.

I've some further refactoring for the build system on my todo (also in connection with Mobile needs and Dashicons), please involve me in the discussion ;)

folletto:
That said, I checked this PR, and I think we can merge. 👍

gwwar:
@sirreal feel free to squash and merge this one.

PR #235, opened 2017-07-06
Build: Fix classname generation

gwwar:
Fixes a regression from https://github.com/Automattic/gridicons/pull/226 where if we called
```jsx
<Gridicon icon="add-image" className="foo" />
```

The SVG would only have class `foo` and would be missing `gridicon` and the icon name.

See related: https://github.com/Automattic/wp-calypso/pull/15758

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "gridicons",
-  "version": "2.0.1-alpha.1",
+  "version": "2.0.1",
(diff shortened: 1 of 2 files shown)


folletto:
Code looks good. :ship:

gwwar:
Thanks @folletto I'll try another update in Calypso next week

PR #237, opened 2017-07-12
gridicons-nametag.svg

mapk:
Adding a new icon called gridicons-nametag.svg.


drw158:
I try to not overlay solid shapes with white strokes, because usually there is a tendency to use a 1px stroke and make the icon looks less solid or too detailed. Do any of these alternate designs on the right look good to you?

Same 3 icons shown at different resolutions

screen shot 2017-07-12 at 12 26 01 pm
screen shot 2017-07-12 at 12 25 53 pm
screen shot 2017-07-12 at 12 25 42 pm

mapk:
Good point, @drw158. I appreciate your help with this!

My concern is that the block formation of the clip element doesn't convey that element very well. What if we did a half stroke?  Something like this?

option-2

mapk:
Or something like this?

option-3

drw158:
Hey sorry, I deleted my comment and moved it to the issue. I'll copy my images over here:

screen shot 2017-07-12 at 12 26 01 pm
screen shot 2017-07-12 at 12 25 53 pm
screen shot 2017-07-12 at 12 25 42 pm

I think I still prefer the 'white' (not filled in) clip over the black. The black clip looks like an 'inverted' color effect.

Maybe something like this to solve the clip problem:

screen shot 2017-07-12 at 1 56 32 pm
screen shot 2017-07-12 at 1 56 24 pm

drw158:
Two more options

screen shot 2017-07-12 at 2 03 49 pm
screen shot 2017-07-12 at 2 03 42 pm

folletto:
screen shot 2017-07-13 at 13 51 39

This would be my prefefence: 

1. It has a black line inside, so it's hinting there's text
2. The pin is clear, and the top extra border makes it stand out better
3. It's the one that looks the least a "box", so less confusion

drw158:
Hope you don't mind, I review the icon and everything looked 👌 

I did combine the icon into one compound shape though and cleaned up the svg output. 

Would you like me to `grunt` and generate all of the files, or do you want to give it a go? I don't mind.

drw158:
I'm not seeing the svg-min icon. Or am I just missing it?

folletto:
You're right, sorry I missed it :|

drw158:
No problem! 

@mapk I saw the svg-min icon in the screenshot you showed me yesterday, but it's a new file, not modified, so maybe it wasn't included.

mapk:
@drw158 I just pushed the branch again with the svg-min icon. Should I create another PR from that?

drw158:
I think that will work

PR #238, opened 2017-07-14
Build: build PDF using `svg-to-pdfkit`

folletto:
screen shot 2017-07-14 at 15 31 13

This PR adds the `svg2pdf` task using [svg-to-pdfkit](https://www.npmjs.com/package/svg-to-pdfkit) (which uses [PDFKit](http://pdfkit.org/)).

The scope of the PDF is to be used in iOS, so we ran a couple of tests with @frosty.

PDF in iOS 10:

pasted image at 2017_07_14 03_14 pm

PDF converted with Paintcode:

pasted image at 2017_07_14 03_26 pm

### To test

* Run `grunt` and make sure everything gets generated as usual
* Try emptying the `pdf` folder and run `grunt` again, all the PDFs should appear
* Check that the PDFs shapes are correct

/cc @mattmiklic 

Props: Massimo Menichinelli @openp2pdesign (tipped me [here](https://www.facebook.com/folletto/posts/10158960582480623?comment_id=10158960761175623&comment_tracking=%7B%22tn%22%3A%22R0%22%7D))


drw158:
Everything worked for me 👍

drw158:
We just need to update the Readme files

folletto:
Thanks!

1. Updated readme
2. Removed `makepdfs.sh` now not relevant anymore

folletto:
I also updated the installation wiki page: https://github.com/Automattic/gridicons/wiki/Installation

I thus noticed that as a nice side effect of this PR the issue mentioned there is solved too:

> **Note**: `cairosvg` outputs its version number in the PDF files it generates. We can't turn this off. But this means if you run the script and suddenly see all the PDFs as changed when running `git status`, this is likely because your version of Cairo was newer or older than the version a previous committer used.

PR #239, opened 2017-07-14
Gridicons nametag

mapk:
Added the svg-min icon and pushed this branch up again. I believe it just needs to be merged now.


drw158:
Looks good but not sure what will happen since the other files are already in master. I'll defer to @folletto !

folletto:
Yeah it would have been easier to take master and add the missing bits. :D 
I did a rebase, and I'm merging just to make sure doesn't break anything as I'm not 100% sure what I did was correct. :D

mapk:
Okay, let me know if I can do anything different to make this easier.

folletto:
It's good, rebase done :)

I've also tweaked a bit the source SVG so it's more readable. :)

PR #241, opened 2017-07-21
Built: refactor tasks to file, demo template

folletto:
screen shot 2017-07-21 at 16 29 35

This PR:

* moves all the custom tasks to individual files so they are more easily reused and updated
* moves the template for the demo outside the Grunt file in its own template file

Why?

1. Easier to spot what changed in the build process
2. More easily reused across different repositories (Gridicons, Social Icons, Dashicons)
3. Cleaner separation of dependencies
4. Folder and files input declared in config instead of being hardcoded
5. Separate demo template
6. Separate react template

Note: there's also one icon added as PDF as it was added before we had the automated PDF build introduced in #238. 

### How to test

1. Switch to this branch
2. `grunt`
3. Check no changes to any built file anywhere, with the exception of the tabulation of the demo.
4. Check the compiled React component works.


folletto:
FWIW, I ran a diff for `index.jsx` and `example.jsx` and they match.

drw158:
everything seems to work 👍

gwwar:
Did we forget to check in pdf/gridicons-nametag.pdf ? I see this as modified when I run npm install.

folletto:
Yes: 704839da512f4003457249af9c346af8abb7359e

gwwar:
For folks 👀 you don't need a global grunt installed, the equivalent script that uses grunt from node_modules is `npm run build`

grunt-tasks/svg-to-pdf.js
```diff
@@ -0,0 +1,30 @@
...
+        pdf.info.CreationDate = "";
+        pdf.pipe( fs.createWriteStream( files.dest + svgFile.slice(0, -4) + '.pdf' ) );
+        pdf.on('close', function() { done(); });
+
+        svg2pdfkit(pdf, fileContent, 0, 0 );
+        pdf.end();
+      } )
+    } );
+  } );
+}
```
gwwar: Don't forget the `;` for assignments. 😄 Looks like the other split up scripts need one too.

folletto:
Ok done:

1. Updated to master
2. Cleaned up the missing ";".

gwwar:
@folletto would you be interested in also updating the social-logos build? We'd be able to get rid of the handlebars dependency then

folletto:
Ok, added to my todo list :)

folletto:
Refactoring of social logos build PR: https://github.com/Automattic/social-logos/pull/58

PR #243, opened 2017-08-14
Update Chat icon

jancavan:
In this PR: https://github.com/Automattic/wp-calypso/pull/17144, we added an Automattic Conversations icon which looks like:

screenshot 2017-08-14 11 28 38

The Conversations icon is currently using the `chat` icon which looks like this:
screenshot 2017-08-14 11 12 06

Both icons don't match very well so we decided to flip the `chat` icon using `transform`. After discussing with @folletto, we decided it's better to just flip the original icon instead of using CSS to do this.

screenshot 2017-08-14 11 12 12

Note: I only updated the icon in `sources/svg`. I didn't run `grunt` yet.


folletto:
It's one of the small tweaks that I feel works well as it would allow more modularity in the future similar to the one you mention.

I also ran a [quick search](https://github.com/Automattic/wp-calypso/search?utf8=%E2%9C%93&q=gridicon+icon%3D%22chat%22&type=) just to be sure, and it doesn't seem there's anything that could break with the change — just standard usage in buttons.

jancavan:
Just ran `grunt`: https://github.com/Automattic/gridicons/pull/243/commits/1bd5d0f4bf7cee1b0b25b27f3623eb37b7499af5

jasmussen:
Change looks good! 👍 👍

PR #246, opened 2017-08-24
New icon: Follow and Following Conversation

jancavan:
In Reader, we're working on a new Conversations Tool: https://github.com/Automattic/wp-calypso/milestone/202 where users can follow/unfollow conversations and we need new icons for it.

Default conversation icon:
screenshot 2017-08-24 15 36 38

Following conversation:
screenshot 2017-08-24 15 37 05


folletto:
The icons are good in general, but in the icon set it doesn't match our current comment icon, nor the chat icon:

screen shot 2017-08-25 at 12 14 52

screen shot 2017-08-25 at 12 15 39

We need these two to be coordinated to the whole set, either by aligning the two suggested here, or changing all of them to be another shape, but all based on the same structure.


_(Btw, that ✓ cut is SHARP, well done)_

jancavan:
> The icons are good in general, but in the icon set it doesn't match our current comment icon

@folletto I gave this a shot, but it looks really weird:
screenshot 2017-08-25 09 49 41

Especially when it's small, it's hard to tell what it is:
screenshot 2017-08-25 09 50 30

> or changing all of them to be another shape

It's better to change the entire bubble set for uniformity, but the "Follow/Following" icons for Conversations will only be available internally for now. Can we get these icons added to the set then I can create a separate PR for the tweaked comment/chat icons?

folletto:
I'd rather not introduce an inconsistency... if we have a change we can change in the same PR?

Have you tried adding the plus in one other corner? Or in negative in the middle?

jancavan:
> Have you tried adding the plus in one other corner?

screenshot 2017-08-25 10 54 20

I tried to add it on the upper right, but we'd really like to make this consistent with the Follow/Following site icon where the plus/check are on the lower right.

folletto:
Yep I agree, it should be consistent everywhere. I'm ok to find it in any direction, as long as it's consistent.

jancavan:
In https://github.com/Automattic/gridicons/pull/246/commits/47f00bdf0040f0619401d37f15636d84b16eaf51, I've updated the 4 icons so they are more uniform with each other. I've also cleaned up some extra nodes in each:

Follow conversation:
screenshot 2017-08-25 19 53 51

Following conversation:
screenshot 2017-08-25 19 54 54

Comment:
screenshot 2017-08-25 19 53 28

Chat:
screenshot 2017-08-25 19 49 24

folletto:
I'm not sure... there's still a change of shape across them that doesn't feel very coordinated, but I see the problem with the "Chat" one in doing that reversal. Also there's a slight movement of the triangle between Follow and Following ones.

Could we try to have at least Follow/Following/Comment to be exactly the same shape?

jancavan:
In https://github.com/Automattic/gridicons/pull/246/commits/5e668b233d5de402e78fa89f4b7734637af43332:

screenshot 2017-08-28 16 51 17

These bubble icons are now using the exact same rounded rectangle and the same pointer/triangle location.

folletto:
Design wise, perfect 💯 

---

Some technical fixes:

(1) The chat one is off-center:

screen shot 2017-08-29 at 10 33 53


(2) The comment one is not centered in a 24x24 square (has `viewBox="0 0 21 18"`):

screen shot 2017-08-29 at 10 34 35

jancavan:
> (1) The chat one is off-center

Corrected in: https://github.com/Automattic/gridicons/pull/246/commits/97c6868a5ec276a7442c85626e947e5cf94d67cc

screenshot 2017-08-29 09 12 45

> The comment one is not centered in a 24x24 square (has viewBox="0 0 21 18"):

As discussed in Slack, this was an unnecessary file that was removed in: https://github.com/Automattic/gridicons/pull/246/commits/cd450e0c35bf876e9b2998819a1f664ce82a848d

folletto:
👍

PR #249, opened 2017-09-05
Adds a half star

raoulwegat:
https://wpcalypso.wordpress.com/devdocs/design/rating  uses noticons for three icons – a filled star, a half star, and an outline star. Gridicons already has a filled star and outline star, but does not have a half-star. To replace the rating component noticons with gridicons, we need a half star. 

screen shot 2017-09-05 at 3 46 55 pm

Calypso rating component: https://github.com/Automattic/wp-calypso/tree/a70587157b33783ba4d5e0e4a1e136680d459d51/client/components/rating 

#hackday


folletto:
We decided in the past it wasn't necessary as any "intermediate" measure could be done with overlaying a full star to an empty star (and that would also allow intermediate grading, like one-fourth and so on.

screen shot 2017-09-05 at 10 51 47

Is there any reason why we can't do it with overlay as before?

folletto:
Btw, with the above I didn't mean to just say "no": if there's a good reason we didn't evaluate for having it, we can review the decision — it's good to ask and review! :) 

I'm sure we discussed it, But... I can't find where... I thought Plugins was already using it but it's using Noticons (whaaa?!? whyyyy)...

jameskoster:
FWIW I also prefer the css approach to this as it allows greater accuracy and fewer icons :)

kellychoffman:
Nice! Whatever way we go with, can we also display it in the dev docs so its obvious we have it?

BTW, this reminds me of the half star bug in etsy story: https://storify.com/ELLIOTTCABLE/four-and-a-horse-stars

raoulwegat:
I'm also in favor of using CSS to create intermediate icons, so I'll close this PR and get to work on:

1. Adding a class to calypso for the half-star; and
2. Update the dev-docs with the new half-star; and
3. Update the docs here to detail how intermediate icons can be handled using the half-star as an example.

Thanks for your input everyone. 😄

folletto:
You're great! Thank you!

PR #247, opened 2017-08-29
Bump to v2.0.3

jancavan:
This PR updates Gridicons to `v2.0.3` and `changelog.md`

PR #250, opened 2017-09-06
Docs: updated NPM details, step by step.

folletto:
More explicit step-by-step explanation to publish to npm.

Questions:

1. Is it thorough? If I know zero apart from initial Gridicons setup and I follow EXACTLY these steps, do I get the right outcome?
2. We mentioned using the npm support for github branches instead of `1.1.0-alpha.1` tag. I feel it's better, but I probably need a review to make sure I'm not writing bad instructions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,7 +87,7 @@ This icon set uses a few automation scripts to ease the generation of new icons 
 Note: to proceed with this you need to have write authorization to npm.
 
 1. Create a new PR with updated `CHANGELOG.md` and updated version in `package.json` (i.e. `1.2.3-alpha.1`), see an example [here](https://github.com/Automattic/gridicons/pull/247).
-2. Pre-publish that PR branch on npm with `npm publish --tag next`.
+2. Pre-publish that PR branch on npm with `npm publish --tag next` [more info](https://docs.npmjs.com/getting-started/using-tags).
 3. Create a new PR in a repository using Gridicons (i.e. [Calypso](https://github.com/Automattic/wp-calypso)), see an example [here](https://github.com/Automattic/wp-calypso/pull/17601).
 4. Test if the new icons show up, and there are no regressions in the previous icons.
 5. If changes look good, remove postfix in the version (i.e. `1.2.3`).
--- a/README.md
+++ b/README.md
@@ -88,7 +88,7 @@ Note: to proceed with this you need to have write authorization to npm.
 
 1. Create a new PR with updated `CHANGELOG.md` and updated version in `package.json` (i.e. `1.2.3-alpha.1`), see an example [here](https://github.com/Automattic/gridicons/pull/247).
 2. Pre-publish that PR branch on npm with `npm publish --tag next` ([more info](https://docs.npmjs.com/cli/dist-tag)).
-3. Create a new update PR in a repository that makes use of Gridicons (i.e. [Calypso](https://github.com/Automattic/wp-calypso)) pointing `package.json` to the alpha version, see an example [here](https://github.com/Automattic/wp-calypso/pull/17601).
+3. Create a new update PR in a repository that makes use of Gridicons and run `npm install gridicons@next --save` (which will update `packages.json`). If you're creating the PR in [Calypso](https://github.com/Automattic/wp-calypso) and you get warnings, it might need to regenerate the shrinkwrap, in which case run `npm run update-deps`. See an example [here](https://github.com/Automattic/wp-calypso/pull/17601).
 4. Test if the new icons show up, and there are no regressions in the previous icons.
 5. If changes look good, remove the alpha postfix in the version (i.e. `1.2.3-alpha.1` to `1.2.3`) on both repositories PRs.
 6. Merge the Gridicons PR.
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ These are not rules, they are guidelines that can be broken with the proper reas
 Note that the icons in this set are tied to be used in [Calypso](https://github.com/Automattic/wp-calypso/), but there might be exceptions for more general icons that make sense to be added.
 
 1. Make sure you have a updated local clone of the repository.
-2. Draw the icon in Illustrator on a 24px grid using the guidelines above (use [icon-template.ai](https://github.com/Automattic/gridicons/wiki/Icon-Template) as starting point).  
+2. Draw the icon in Illustrator on a 24px grid using the guidelines above (use [icon-template.ai](https://github.com/Automattic/gridicons/wiki/Icon-Template) as starting point).
    _Tip: tap CMD + Option + Y in Illustrator to see the pixel grid version._
 3. Submit a Pull Request with the icon as a SVG file (inside the `sources/svg` folder), make sure to include a screenshot, ideally containing side by side comparison with some other Gridicons as a visual reference.
 4. Discuss, iterate, review, refine until ready.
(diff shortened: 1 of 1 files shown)


folletto:
Current PR text:


## Publishing to npm

Note: to proceed with this you need to have write authorization to npm.

1. Create a new PR with updated `CHANGELOG.md` and updated version in `package.json` (i.e. `1.1.0-alpha.1`), see an example [here](https://github.com/Automattic/gridicons/pull/247).
2. Pre-publish that PR branch on npm with `npm publish --tag next`.
3. Create a new PR in a repository using Gridicons (i.e. [Calypso](https://github.com/Automattic/wp-calypso)), see an example [here](https://github.com/Automattic/wp-calypso/pull/17601).
4. Test if the new icons show up, and there are no regressions in the previous icons.
5. If changes look good remove postfix in the version (i.e. `1.1.0`).
6. Merge the Gridicons PR.
7. Publish MASTER using the latest tag `npm publish --tag latest`.
8. Merge the test repository PR.

gwwar:
Thanks @folletto! 

@jancavan how do you feel about these instructions? Do you feel like this is enough information for you to publish another gridicons version?

gwwar:
> We mentioned using the npm support for github branches

I tried that and `npm link` in the past, but it doesn't work well for Gridicons since we have special build steps. If folks see otherwise, we'd be happy to take other the dev flow improvements.

jancavan:
> 7. Publish MASTER using the latest tag npm publish --tag latest.

IIRC, I had to do these after the step 7:

`git tag -a v2.0.3 -m "v2.0.3"`
`git push origin v2.0.3`

Then checked if the new version is now showing up on: https://github.com/Automattic/gridicons/releases

jameskoster:
I think we could explain the `npm publish --tag next` and `npm publish --tag latest` commands. 

I'm new to npm tagging and it's not 100% clear what's happening here. I understand what we're doing and why we're doing it, but the tag names in the example have me a little confused. Are `next` and `latest` the literal tag names?

If so, wouldn't it be better to use something more explicit than `next` for a tag name? Something like `1.1.0-alpha` for example? Otherwise won't we have a bunch of duplicate `next` tags?

Linking to the example PR in step 3 could perhaps be expanded as well. Obviously that's the commit that was actually merged into master, but beforehand, to test, you have to run something like `npm install gridicons@next`, right? That step seems to be missing.

folletto:
> I tried that and `npm link` in the past, but it doesn't work well for Gridicons since we have special build steps. If folks see otherwise, we'd be happy to take any dev flow improvements.

What are the extra build steps?
Shouldn't they be included in the list above then?

And... is `npm link` the same as specifying the repository in `package.json` ([ref](https://docs.npmjs.com/files/package.json#git-urls-as-dependencies))?

What I was hinting was to do this instead in the PR branch when testing (or local testing before publishing to npm):

```
    "gridicons": "Automattic/gridicons#branch\/name",
```

I'm mostly trying to avoid the double-update of `v1.1.1-alpha.1` and then `v1.1.1`. If there's a way we can just test it without the extra step, isn't it better?

folletto:
> IIRC, I had to do these after the step 7:

Thanks, added to the readme! :)

folletto:
> Are next and latest the literal tag names?

Afaik, no: `next` is a convention ("The next tag is used by some projects to identify the upcoming version.", [source](https://docs.npmjs.com/cli/dist-tag)), and `latest` is special (see [here](https://docs.npmjs.com/cli/publish)).

> I'm new to npm tagging and it's not 100% clear what's happening here. 

I think the checklist should be strictly operative, so no explanations. We can add link for details if someone is curious tho. :) 

---

Actually... As `latest` is default, is there any issue in just changing the command to `npm publish`?

jancavan:
I made some minor changes in: https://github.com/Automattic/gridicons/pull/250/commits/94818d1c38b3d15b74601b91467c3fc312874440

Btw, I don't remember having to do the postfix part (`-alpha.1`). I only ever updated the version number and changelog. Is this step necessary?

gwwar:
> Btw, I don't remember having to do the postfix part (-alpha.1). I only ever updated the version number and changelog. Is this step necessary?

@jancavan It's if you want to test changes before finalizing the version.

@folletto npm link uses your local copy, a github url will download from a branch/hash.

https://docs.npmjs.com/cli/link

> What I was hinting was to do this instead in the PR branch when testing (or local testing before publishing to npm):

Yes, you're welcome to try to verify, but I found that both approaches doesn't work with this package. It might be due to us not checking in the build folder. If you find an approach that works, feel free to update steps here.

Another way to avoid alpha packages would be to add CI and component tests.

jameskoster:
> I think the checklist should be strictly operative, so no explanations. We can add link for details if someone is curious tho. :)

Fair point, I agree.

Once this ready to merge I can use it to do my first deploy as a test run and provide further feedback.

folletto:
Ok updated:

1. Link to "more info" (fixed the formatting too)
2. Simplified to `npm publish` so to avoid ambiguity with tags


> Yes, you're welcome to try to verify, but I found that both approaches don't work with this package. It might be due to us not checking in the build folder. If you find an approach that works, feel free to update steps here.

Ok, I'll note then to do this after this PR gets merged as it would require more review.

> It's if you want to test changes before finalizing the version.

Now that I think of it... it's not used anywhere in the steps above. If it's to test, would that mean that after going with `-tag next` the Calypso PR should include the `v1.2.3-alpha.1` version, test, THEN change to `v1.2.3`?

gwwar:
> Calypso PR should include the v1.2.3-alpha.1 version, test, THEN change to v1.2.3?

Exactly, and in the related Calypso PR, I'll put a "DO NOT MERGE" in the summary until the package is finalized.

folletto:
Ok, I updated the README to clarify the alpha steps. :)

README.md
```diff
@@ -88,14 +88,14 @@ Note: to proceed with this you need to have write authorization to npm.
...
+3. Create a new update PR in a repository that makes use of Gridicons (i.e. [Calypso](https://github.com/Automattic/wp-calypso)) pointing `package.json` to the alpha version, see an example [here](https://github.com/Automattic/wp-calypso/pull/17601).
 4. Test if the new icons show up, and there are no regressions in the previous icons.
-5. If changes look good, remove postfix in the version (i.e. `1.2.3`).
+5. If changes look good, remove the alpha postfix in the version (i.e. `1.2.3-alpha.1` to `1.2.3`) on both repositories PRs.
 6. Merge the Gridicons PR.
 7. Tag the release on GitHub: `git tag -a v1.2.3 -m "Release v1.2.3"` (and push `git push origin v1.2.3`).
 8. Check if it shows up in the [Releases list](https://github.com/Automattic/gridicons/releases).
 8. Publish to MASTER using the latest tag `npm publish`.
-9. Merge the test repository PR.
+9. Merge the update PR in the other repository.
```
jancavan: Updated?

folletto: It's meant to match point 3: "Create a new update PR in a repository...", but probably there's a better way to use the same wording on 3 and 9 so the reader know it's the same PR?

jameskoster:
OK I just worked through this to do a release. It went fairly smoothly. Step 3 could be clearer though I think. Updating package.json in Calypso to point at the beta tag then running `npm install` didn't do anything for me. Instead I ran `npm install gridicons@next` to test.

folletto:
> Updating package.json in Calypso to point at the beta tag then running npm install didn't do anything for me. Instead I ran npm install gridicons@next to test.

Interesting. I might be mistaken, but `npm install something` overrides the local `package.json` specification with the one you installed manually. So maybe there's a different way to do it via `package.json` that works?

What's your take @gwwar?

gwwar:
`npm install something` without a tag is equivalent to `npm install something@latest`

folletto:
Aha so `npm install gridicons@next` updates `packages.json` too? Neat. 

So what's the best option in this case among the two?

Would this be ideal?

> 3. Create a new update PR in a repository that makes use of Gridicons (i.e. [Calypso](https://github.com/Automattic/wp-calypso)) and run `npm install gridicons@next` (which will update `packages.json`), see an example [here](https://github.com/Automattic/wp-calypso/pull/17601).

gwwar:
Sorry to clarify `npm install gridicons@next` will update your local node_modules

`npm install gridicons@next --save` will attempt to update package.json and the shrinkwrap the latter may fail (with warnings) if that happens call `npm run update-deps` to regenerate the shrinkwrap in Calypso

folletto:
Ok so would this work?

> 3. Create a new update PR in a repository that makes use of Gridicons (i.e. [Calypso](https://github.com/Automattic/wp-calypso)) and run `npm install gridicons@next --save` (which will update `packages.json`). It might need to regenerate the shrinkwrap, in which case run `npm run update-deps`. See an example [here](https://github.com/Automattic/wp-calypso/pull/17601).

gwwar:
Yes, that should work. We'll want to clarify that `npm run update-deps` is a Calypso only workflow.

folletto:
Ok, iterating with that:

> 3. Create a new update PR in a repository that makes use of Gridicons and run `npm install gridicons@next --save` (which will update `packages.json`). If you're creating the PR in [Calypso](https://github.com/Automattic/wp-calypso) and you get warnings, it might need to regenerate the shrinkwrap, in which case run `npm run update-deps`. See an example [here](https://github.com/Automattic/wp-calypso/pull/17601).

jancavan:
I'll have to try these steps out again next time because I remember updating `packages.json` manually.

folletto:
Yeah makes sense, we need to run this again to be sure. :)

Meanwhile, I updated step 3 in the doc. I'd say we can merge this now, and iterate on it later? Can I get one thumbs up? :D

PR #253, opened 2017-10-05
Readme: Clarify usage outside a React project

iandunn:
If you don't read the previous note closely, it'd be easy to think that using Gridicons requires using React. This makes it obvious that it's easy to use Gridicons in non-React projects.


folletto:
Ah yes! Excellent point, and while I think we could do even more, your suggestion is already an excellent improvement. Thanks!

Extra thanks for fixing the double dot :P

PR #254, opened 2017-10-09
New/Update icons: image, video, and remove variants

iamthomasbishop:
This introduces 2 new icons: "remove image" and "remove video". These are meant to be used to indicate the user is removing an image or video from the editor canvas, to distinguish from the "delete" action (Trash Can icon).

proposed


folletto:
They both look great! 

I've however a concern on how the "slash" goes further beyond the limits of the template: the picture and video images were already using the extra width due to the lack of height, which is ok, but that is already the "extra" space as the icon should sit inside the inner square.

Which I think it's visible side by side with the eye, that seems optically smaller now.

I'm not sure what's the solution here, as this has to match the original icons too. Maybe we should redesign all the icons to be smaller? But then the original one would feel smaller, while the slashed one would feel good. 

Ideas?

jameskoster:
Totally debatable but I'd consider removing the little notches highlighted in this img;

notches

Details this small are probably going to get lost or add weirdness at smaller sizes.

iamthomasbishop:
@folletto Good point. What do you think of something like this, where the underlying icons are narrower (square on the grid)?

proposed-v2

This seems a bit better and more consistent to me. Also cleaned up the detail @jameskoster mentioned above.

If this is the right direction, I'll just need to clean up some of the lines slightly.

mattmiklic:
> What do you think of something like this, where the underlying icons are narrower (square on the grid)?

These look great, particularly at smaller sizes. I like the way the notches on the film icon are cleaner in this version. 

The only thing I wondered about whether this and the standard image icon would ever appear next to each other, or would alternate with each other in the same position. If that happened, the difference in widths might be noticeable. But I don't think we'd be using this icon in that way, so it's probably not a concern.

folletto:
Super! 

> The only thing I wondered about whether this and the standard image icon would ever appear next to each other, or would alternate with each other in the same position. If that happened, the difference in widths might be noticeable. 

Wondering the same.

Is there any drawback we can think of in aligning all the icons with the above? I think will make the original "image" and "video" ones more modular for other uses.

jameskoster:
> Is there any drawback we can think of in aligning all the icons with the above? I think will make the original "image" and "video" ones more modular for other uses.

I think that's worth exploring. We already have `image-multiple` which uses the smaller, square photo glyph so we'd be adding consistency there. \

It would be cool to see these new icons along side their counterparts (and some other Gridicons as well). Maybe in context of the UI as well.

iamthomasbishop:
Updated the lines on the image-remove icon to be cleaner, more balanced:

icon-diff

folletto:
> We already have image-multiple which uses the smaller, square photo glyph so we'd be adding consistency there.

I forgot this, excellent! Then... I'd suggest to realign all the icons to the square-image... :P 

> Updated the lines on the image-remove icon to be cleaner, more balanced:

Superb detail attention 💯

iamthomasbishop:
> Superb detail attention 💯

Glad you approve, @folletto  :)

If there are no other objections or concerns with these icons, would you mind a final review and merge? // cc @jameskoster

folletto:
> If there are no other objections or concerns with these icons, would you mind a final review and merge?

My only concern is having this not aligned. 
Can we change all the media icons at once, so both "slashed" and normal? Do you prefer two separate PRs?

iamthomasbishop:
> Can we change all the media icons at once, so both "slashed" and normal? Do you prefer two separate PRs?

@folletto My original thought was to get these into the mix so they'd be available for use in the Media Settings project, but I'll go ahead and update the base image/video icons as well and push shortly.

iamthomasbishop:
So just to confirm, the icons that need updates to be squared off are:

* gridicons-add-image
* gridicons-image
* gridicons-video
* gridicons-image-multiple

...and the ones already updated:

* gridicons-image-remove
* gridicons-video-remove

Am I missing any? The video ones are pretty straight forward, but the image ones are a little trickier to grok.

folletto:
It seems to me `gridicons-add-image` and `gridicons-image-multiple` should already be squared, but worth to double check. :)

I think that's the case, yes. :)

iamthomasbishop:
@folletto They are squared, but I was thinking we might want to clean up the lines a bit to match these new ones. Overkill? Save it for later?

folletto:
If I was doing me I'd try to do it now just to take the whole thing off — but if time is a variable, I would postpone it. :)

iamthomasbishop:
I'm thinking after all that it might be best to break the additional icon updates into a separate PR as I haven't gotten to the other ones quite yet and would like to provide them for use right away in the updated media settings screens on the native apps. Because we aren't using these `remove` icons in any other places, it shouldn't be a blocker, and will give us time to update the others.

I updated the base image icon on this `gridicons-image-remove` icon (which will help clean it up for use in the forthcoming base `gridirons-image` icon) and pushed a commit. Here's what it looks like:

screenshot 2017-10-17 15 07 55

Note: `gridicons-video-remove.svg` remains the same as the previous iteration.

Sound good, @folletto? If so, mind taking a look and approving/merging?

folletto:
I did some cleanup to the SVG:

1. Made it a single path
2. Remove all `translate` rules
3. Remove all irrelevant IDs
4. Remove all `style` and other markup that wasn't useful
5. Size set at 24px

Have a look at the source diff between before and after to get an idea. It's a mix of work in Sketch to cleanup and merge paths (I can't do it manually), removing groups and so on, and some cleanup manually afterward. 

There are no rules yet to this, so I've done it as an exercise to understand better what we might want when we merge SVGs.

I also added the `gridicon-image` squared SVG as I was there.

screen shot 2017-10-18 at 12 22 09

screen shot 2017-10-18 at 12 22 17

folletto:
Ok, I did the same for video:

screen shot 2017-10-18 at 12 44 05

screen shot 2017-10-18 at 12 44 14

folletto:
Test it a bit: both in browser and dropping it into image editing apps. If all it's ok you can go forward by [building and merging this](https://github.com/Automattic/gridicons#add-a-proposed-icon-to-gridicons-admins-only). :)

PR #255, opened 2017-10-13
Gridicons: Rotate Reply icon

jancavan:
The Reply icon is inconsistent in two places: My Sites > Comments and Reader.

The original gridicon looks like this:

screenshot 2017-10-13 15 27 54

Which is what's being used in Comments:
screenshot 2017-10-13 15 42 21

But in Reader, it's being flipped (using CSS):
screenshot 2017-10-13 16 00 18

I'd like to propose that we flip the original icon so it's consistent everywhere, like so:

screenshot 2017-10-13 15 28 00

And once this PR is approved, I can go ahead and un-flip the Reader version.


folletto:
Now I understand, it was meant to be used in the other direction the whole time :D 

Looks good to me, just a simple flip — and it also make more sense in terms of direction 👍

folletto:
I did some cleanup on the code too. 

Can you double check if it works for you before merging? Just to be sure :)

jancavan:
Thanks for cleaning up that SVG @folletto! Looks good 👍

Issue #242, opened 2017-07-28
Update icon: Create (Write)


folletto:
![screen shot 2017-07-28 at 15 38 54](https://user-images.githubusercontent.com/4389/28724131-f2114c7a-73b0-11e7-92f1-e333cba66f09.png)

![screen shot 2017-07-28 at 15 37 20](https://user-images.githubusercontent.com/4389/28724137-f5ac1978-73b0-11e7-8954-4d3aab200f60.png)

We had some feedback on mobile that the "Write" icon (`gridicon-create`) is a bit too obscure. This might be due to various reason, but we could also take the chance to resolve the "border inside border" effect we currently have, as well as creating something more distinctive (ref: Twitter).

`gridicon-create`:

![screen shot 2017-07-28 at 16 20 49](https://user-images.githubusercontent.com/4389/28724149-fee2f016-73b0-11e7-809a-07be7976148b.png)

For reference, before we used `gridicon-pencil`, which we use now for generic edit:

![screen shot 2017-07-28 at 16 20 30](https://user-images.githubusercontent.com/4389/28724142-f8cd93e8-73b0-11e7-9eba-78f7b7c3d125.png)

Ideally we'd like something clearer as well as more unique.

drw158:
Previous work: #84

drw158:
Just some thoughts:

It's hard to imagine something more clear for 'Write' than the pencil icon. The 'pencil-in-a-box' icon is also pretty standard from my POV, but I realize I'm biased in that way. It's possible that the Write icon is not even clear that it's a pencil, which would be a problem! I like the idea of just a pencil. Some ideas I would try:

- make it more clear that it's a pencil
- try to make the 'box' look like paper
- try a pencil icon that's drawing a line

Some icon ideas I like:

<img width="164" alt="screen shot 2017-07-28 at 10 50 06 am" src="https://user-images.githubusercontent.com/618551/28725578-7ca8d26e-7383-11e7-904e-ae8d1cf3c746.png">
<img width="152" alt="screen shot 2017-07-28 at 10 49 51 am" src="https://user-images.githubusercontent.com/618551/28725579-7caeab58-7383-11e7-94e8-156ab9252b14.png">
<img width="28" alt="screen shot 2017-07-28 at 10 49 28 am" src="https://user-images.githubusercontent.com/618551/28725580-7cb29b3c-7383-11e7-810d-f30b17505c71.png">
[...]

PR #256, opened 2017-10-16
Update icon: create

folletto:
screen shot 2017-10-16 at 13 55 40

screen shot 2017-10-16 at 13 58 01

This icon proposal builds on the long discussion on #242. 
As stated there this change will improve under these aspects:

1. To make the icon more understandable, as the pencil/paper abstract one didn't perform too well and plus seems a better choice to "add content"
2. To make the icon more generic so it would suggest not just text but also images.
3. Being an incremental change on the previous one.
4. Still being different from the other "plus" icons we have.

Downside:

1. It de-emphasise the "text" component of creating new content

As the discussion happened already extensively in #242 for 2+ months, the feedback at this stage is on the technical aspects of it:

* SVG cleanliness
* Does it work in Sketch / Illustrator
* Does it conform to the Gridicons standards

If yes, thumb up and we'll merge. ;)


jasmussen:
Visually I love this one. 👍 👍

I had some npm woes with this one, not sure why — probably issues on my end:

screen shot 2017-10-23 at 12 58 00

folletto:
Odd, the SVG source got mangled in the build process:

```
<svg viewBox="0 0 2<g><g>4 24" xmlns="http://www.w3.org/2000/svg"><path d="M21 14v5c0 1.105-.895 2-2 2H5c-1.105 0-2-.895-2-2V5c0-1.105.895-2 2-2h5v2H5v14h14v-5h2z"/><path d="M21 7h-4V3h-2v4h-4v2h4v4h2V9h4"/></g></g></svg>
```

Note the `<g><g>` in the middle. I feel something's broken with the `<g>` insertion. Looking into that.

jasmussen:
I think I've seen this. Try and outline/flatten the icon in question. I think that has to happen regardless, because the React component can only handle a single path inside the file (which is a problem probably worth addressing separately).

folletto:
Ah possibly, I was trying to keep the plus separate. I'll check that too.

folletto:
Found: `svg-transform-add-g` specifically checks `viewBox` at the END of the SVG. Looking at ways to refactor it.

folletto:
Ok, the new build task in #259 should fix it. :)

drw158:
Looks good! 👍

folletto:
Wonderful. This should automatically land in Calypso the next time we update Gridicons. 

If we want this to happen sooner, we can push it. :D

PR #258, opened 2017-10-20
Build: update svg-to-pdfkit to 0.1.3 (exact)

folletto:
During #254 we noticed that when @iamthomasbishop tried to build, he got also all the PDF re-generated as different. This shouldn't happen — so after fixing the PR we tried to identify the issue.

Apparently, [SVG-to-PDFKit](https://github.com/alafr/SVG-to-PDFKit) updated to 0.1.3, thus changing some calculations on how the PDF was generated.

As it seems it's not a deterministic calculation, this PR:

1. Updates  `svg-to-pdfkit` to 0.1.3
2. Fixes the `svg-to-pdfkit` library to a specific version, so when we decide to update it's manual and won't risk confusing PRs.
3. Updates all the PDFs to use 0.1.3 generated version.

### To test

1. Switch to this branch
2. Try `npm install` then `grunt`
3. Verify with `npm ls pdfkit` that the library version is `0.1.3` and pdfkit is `0.8.3`
4. Make sure no change happens on the generated files

_(Note: the "react" error in `npm ls` is not an issue)._


folletto:
Difference in PDF files:

screen shot 2017-10-20 at 17 49 59

Note that there was no rendered difference in the PDF files... still, worth checking. ;)

iamthomasbishop:
> Verify with npm ls pdfkit that the library version is 0.1.3 and pdfkit is 0.8.3

Looks right to me. Also seeing no changes on the generated files. Thanks for taking care of this @folletto !

PR #260, opened 2017-10-24
React: Allow v16

ockham:
Only bumping the version to 2.1.0(-alpha.1, for now) since we're still permitting `^15.3.0`, too.

Relevant for https://github.com/Automattic/wp-calypso/pull/19083

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "gridicons",
-  "version": "2.1.0-alpha.2",
+  "version": "2.1.0",
   "main": "build/index.js",
   "scripts": {
     "build": "grunt --verbose",
--- a/package.json
+++ b/package.json
@@ -1,13 +1,13 @@
 {
   "name": "gridicons",
-  "version": "2.0.5",
+  "version": "2.1.0-alpha.1",
   "main": "build/index.js",
   "scripts": {
     "build": "grunt --verbose",
     "prepublish": "npm run build"
   },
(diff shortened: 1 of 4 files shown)


sources/react/example-footer.jsx
```diff
@@ -1,4 +1,6 @@
 			</div>
 		);
 	}
-} );
+};
```
gwwar: The semicolon here isn't needed

gwwar:
The generated example file looks good, was there a Calypso PR to test with?

ockham:
> The generated example file looks good, was there a Calypso PR to test with?

Not yet, since I couldn't publish this one. Can you add me to the npm package?

sources/react/example-footer.jsx
ockham: 41c0f51

gwwar:
> Can you add me to the npm package?

Done!

package.json
```diff
@@ -1,13 +1,13 @@
 {
   "name": "gridicons",
-  "version": "2.0.5",
+  "version": "2.1.0-alpha.2",
```
gwwar: Bring this to 2.1.0 before merge and we should be all set.

ockham:
Thanks for reviewing @gwwar and @samouri!

samouri:
we also need to remove `PropTypes` from gridicons

ockham:
Hmm, why is that?

PR #263, opened 2017-12-05
Add Zoom in/out icons

jancavan:
We would like to be able to expand/shrink images in Reader: https://github.com/Automattic/wp-calypso/issues/20478

I couldn't find alternatives to use from the existing set so this PR proposes the addition of zoom in/out icons:

screenshot 2017-12-05 11 24 15

Example usage:

(Zoom in):
screenshot 2017-12-05 13 27 13

(Zoom out):
screenshot 2017-12-05 13 27 07

I tried it with the handle on the right to be consistent with the Search icon, but for some reason, it doesn't feel right:

screenshot 2017-12-05 13 29 23

So for consistency, I think the Search icon should be flipped:

Before:
screenshot 2017-12-05 13 30 02

After:
screenshot 2017-12-05 13 30 10


drw158:
Hm I think the icon just looks a bit awkward because it's in the top right corner of the image; it would look different if it was in the top left corner. Either way, both look fine to me though. You might alleviate the awkwardness by putting the icon inside a button.

Doing a quick icon search, our current icon's orientation seems to be following the most common way, including the Material Design icons. I think we should keep the current orientation.

When used in the search bar, the new orientation you propose looks backward to me, but that might be just me getting use to it :)

jancavan:
> Hm I think the icon just looks a bit awkward because it's in the top right corner of the image; it would look different if it was in the top left corner.

I'm okay with keeping the orientation of the current Search icon if it matches current standards. Do you think it's too weird though? Maybe it'll eventually grow on me.

screenshot 2017-12-05 13 29 23

folletto:
For me the `\` orientation is good :D 

Feels nice to me as it's almost like when you fold the corner of a piece of paper, it's aligned nicely like that. :)

jancavan:
In https://github.com/Automattic/gridicons/pull/263/commits/5a6167f73e8d16ab1042abcffde19a1c11abcf03, I've updated the zoom in/out icons so their handles are on the right:

screen shot 2017-12-07 at 12 27 54 pm

jancavan:
Can I get another 👀  again @drw158 @folletto?

drw158:
Looks like the icon template .ai file has been modified; we shouldn't modify that one. Also, I think because you flipped the search icon back to where it was, there are changes being shown, where there shouldn't bee any changes in the svg code.

Thanks for adding these icons!

jancavan:
@drw158 Thanks for the 👀 . Closing this and creating a new branch instead: https://github.com/Automattic/gridicons/pull/264

PR #264, opened 2017-12-08
Add Zoom In and Zoom Out icons

jancavan:
I've inadvertently updated files that I shouldn't be updating in: https://github.com/Automattic/gridicons/pull/263 so I decided to just create a new branch that supersedes that one.


folletto:
Looks good!

I'd just do some minor cleanup in the SVG, I started a draft here if you want to give it a go:
https://github.com/Automattic/gridicons/wiki/Guidelines-for-SVG-Sources

Regardless, you can build and merge. 👍

jancavan:
Thanks @folletto. I cleaned them up in: https://github.com/Automattic/gridicons/pull/264/commits/951461b79d1fca774f5e153f797af945d908cdd1.

Should the `fill` attribute not be removed? It seems like so in the instructions, but it still exists in the example: https://github.com/Automattic/gridicons/blob/b5d36227e554502e78c403c27e6779fc46c470fd/sources/svg/gridicons-create.svg?short_path=34ad600#L3.

folletto:
> Should the fill attribute not be removed?

Maybe it's not clear in the instructions ("fill attributes excepted to make the icon readable, they will get removed in the grunt process")? 

What I meant is that `fill` is an exception to make the SVG visible. Otherwise would be unfilled and invisible when reviewing the icon. It's still worth to check all the fills to make sure there aren't too many of them, or using odd colors (it should be always `#00000`).

folletto:
SVG code checked: looks nice and clean to me! 💯 

You can proceed building and merging! 👍

jancavan:
Oh I see, makes sense now. So it's okay to keep the `fill` attribute to make make the icon visible in the reviewing process since it'll be removed once you run `grunt`. Gotcha 👍  Thanks for the review!

folletto:
Something odd happened here: there shouldn't be all the PDFs changed.

folletto:
Could you try to see if it's related to https://github.com/Automattic/gridicons/pull/258 by opening a command line and typing `npm ls pdfkit` in the Gridicons folder?

jancavan:
This is what I see @folletto:

```gridicons@2.1.1 /Users/jancavan/gridicons
├── UNMET PEER DEPENDENCY react@^15.3.0 || ^16.0.0 
└─┬ svg-to-pdfkit@0.1.2 invalid
  └── pdfkit@0.8.3  extraneous

npm ERR! peer dep missing: react@^15.3.0 || ^16.0.0 , required by gridicons@2.1.1
npm ERR! invalid: svg-to-pdfkit@0.1.2 /Users/jancavan/gridicons/node_modules/svg-to-pdfkit
npm ERR! extraneous: pdfkit@0.8.3 /Users/jancavan/gridicons/node_modules/pdfkit```

jancavan:
Sorry, I thought that was pretty standard because when I run `grunt`, a lot of files get generated.

folletto:
Ok, it was using the old version sadly :(

As you can see it's something I fixed in https://github.com/Automattic/gridicons/pull/258 but I forgot to think about the case where someone was already having it installed... :/

Could you update with `npm install` and then do a PR to restore the updated PDFs? If you don't have time I'll do it tomorrow (it's a bit late here now).

PR #267, opened 2017-12-11
Revert "Add Zoom In and Zoom Out icons"

jancavan:
Reverts Automattic/gridicons#264


jancavan:
@folletto I'm going to just revert the old branch and do an `npm install`.

folletto:
Ok, let's fix this quickly! ;)

PR #273, opened 2018-01-10
Add plans icon

mattmiklic:
Fixes #240, #261. Props to @MichaelArestad for the 24px hinted version of the Jetpack logo.

screen shot 2018-01-10 at 2 21 33 pm

screen shot 2018-01-10 at 2 21 16 pm

PR #274, opened 2018-01-11
Fix: turns out the rect in the source breaks things

folletto:
I expected adding the rect in the original SVG to not have impact, but it gets read as default black once the whole package is built.

This PR removes the `rect` in the source SVG.

This fixes #273's icon in the build.


folletto:
I also updated the SVG wiki instructions [here](https://github.com/Automattic/gridicons/wiki/Guidelines-for-SVG-Sources).

PR #279, opened 2018-02-02
Add Free and Google image icons

jancavan:
In the Editor, we're currently using the same icon (`gridicons-add-image.svg`) for both "Media from Google" and "Free photo library."

This PR proposes to add two different icons and use these instead so they are more distinguishable from "Media".

Media from Google:
screenshot 2018-02-02 15 26 09

Free photo library:
screenshot 2018-02-02 16 54 25

**Before:**
screenshot 2018-02-02 15 25 39

**After:**
screenshot 2018-02-02 16 54 30

Sidenote:

The "Add Media" icon seems to be flushed right in the grid. Is this intentional?

screenshot 2018-02-02 15 27 42

/cc @jasmussen @mattmiklic @johnmaeda @drw158


folletto:
> The "Add Media" icon seems to be flushed right in the grid. Is this intentional?

It's intentional to keep optical alignment: otherwise the icon, being the bulk a filled square, would feel smaller and misaligned when side-by-side to the others.

folletto:
My take on this change:

1. I think we can drop the special usage of the "+" here: it's already inside a "+" menu, there's imho no need to add a "+" again to the icon.
2. Following from this, we can probably swap out the existing `gridicons-add-image` and use just `gridicons-image`
2. Following from this, I'd just use the default Google logo from Automattic/social-logos for Google
3. I'd also avoid text inside icons, as... well text is already the label. If we want to differentiate the icon, we should find a way to visualize it that isn't by adding "Free" as text. An idea would be to have the mountain and sun hollow instead of filled.

jasmussen:
> Following from this, I'd just use the default Google logo from Automattic/social-logos for Google

I think we should avoid using the actual Google logo, given we're using the Picasa API. Perhaps we can use the actual [Picasa logo](https://venturebeat.com/wp-content/uploads/2016/02/picasa_logo.png?fit=780%2C410&strip=all)? We should probably still call it "Photos from Google".

drw158:
My first thought is that the Google icon neither says "Google" or "media" to me. We should avoid using words in the icon for the free photos icon.

I also like the idea of using the Google logo. It would be great if we could bring back the Google Photos logo. It would more clear that the photos are from _your_ library, not a Google image search or something like that.

> I think we should avoid using the actual Google logo, given we're using the Picasa API. Perhaps we can use the actual Picasa logo? We should probably still call it "Photos from Google".

I do like the Picasa logo, but wouldn't that be confusing? Everything is labeled "Google" from the connection setup to the text in the UI. From a user pov, it's just displaying my Google photos. "I don't know what Picasa is."

If we keep the text label "Google" instead of "Picasa" then perhaps we should just create a generic shutter Gridicon instead of using the Picasa logo.

My suggestions/ideas:

- "Media"— Regular image icon, WordPress icon, new 'media' icon that contains audio + image 
- "Google Photos library" (Media from Google)— Google logo, Google Photos logo
- "Free photo library"— Regular image icon

---

On the subject I found some inconsistencies with the media popover:

- different icons for the google photos
- "Media from Google" vs "Photos from Your Google library"
- "Your" is capitalized

media popup|add button
---|---
screen shot 2018-02-05 at 11 28 56 am|screen shot 2018-02-05 at 11 29 10 am

jancavan:
> I think we can drop the special usage of the "+" here: it's already inside a "+" menu, there's imho no need to add a "+" again to the icon.

I like this idea. The contact form and payment options don't have a plus sign either.

> It would be great if we could bring back the Google Photos logo.

Was this something we used before and just got rid of @drw158?

jancavan:
Another thing I realized - once you connect your Google account, I don't know how to unlink it to try reconnecting to another account.

jancavan:
Ok, so let's go with the Picasa logo for "Media from Google":

screenshot 2018-02-06 11 15 38

What do you think about something like this for "Media"?
screenshot 2018-02-06 11 15 52

Then we can use `gridicons-image` for "Free Photo Library."

Another alternative would be to use `gridicons-image` for "Media" and `gridicons-image-multiple` for "Free Photo Library."

folletto:
Your proposal looks good to me! 👍

jancavan:
This is essentially what it would look like in the Editor's "Add" dropdown:

screenshot 2018-02-06 11 17 02

jancavan:
Question: Should the `gridicons-add-image.svg` icon be removed from the set since I don't think it's being used elsewhere? I don't know if we need to keep unused icons "just in case for future use."

jancavan:
@folletto Should the "Add Image" icon be removed? I don't think it's being used elsewhere?

screenshot 2018-02-09 09 59 42

drw158:
The Add Image icon is being used in /media, and I think in the apps too. I also think it's generic and useful enough to keep in the repo.

I actually prefer this suggestion over the new media icon:

> Another alternative would be to use gridicons-image for "Media" and gridicons-image-multiple for "Free Photo Library."

jancavan:
They look much more coordinated, at least size-wise:

screenshot 2018-02-09 10 16 42

drw158:
I prefer that, but I still see the merit in exploring a new media icon, perhaps in another issue.

folletto:
> They look much more coordinated, at least size-wise:

Looks good! Let's go for it! 👍

jancavan:
I have removed the proposed Media icon so we're adding just the "Media from Google" icon in this PR. I ran `grunt` to generate all the necessary files. 

Can I please get one more 👀 before merging @folletto?

drw158:
Do y'all think this is best suited for Gridicons or Social Logos? I could see it going either way.

folletto:
It feels to me generic enough to go in Gridicons. Maybe we want to make sure it's an actually different shape (but same concept) than Picasa's so we're 100% sure?

jancavan:
I thought we wanted to use the Picasa logo to somehow represent that it is "Google photos"?

@jasmussen What are your thoughts?

folletto:
Yes, true. 
My rationale is that I don't think many users will recognize that as being Picasa's regardless. So we can avoid using it and using instead a "generic" shutter (a similar one is ok) and include in Gridicons available for other uses too. :)

jasmussen:
I like using the Picasa logo, but in my understanding Davide was suggesting that the picasa logo in black and white is "generic" enough to just be a camera shutter icon. 

Either is fine with me. 

I do think logos should ideally be in social logos.

jancavan:
I agree on it not being recognizable. I think for this reason, we can just add it to Gridicons. In the event that we will need to change it to use a different icon, we still have a generic shutter icon to use for something else.

What do you think about any of these?

screenshot 2018-02-20 12 54 06

The difference between (3) and (4) are really just thicker lines which works out better at a smaller size.

screenshot 2018-02-20 13 14 18

drw158:
These look great! I like 4 or 2. A part of me likes the straight lines, but 4 is a nicer icon overall. The white gaps should be 2dp, but have have made some exceptions. Thank you for the mockups.

jancavan:
Thanks @drw158. Yes, my preference is also (4).

folletto:
I agree on 4 — I think it also show more crisply in a small format. 👍

PR #280, opened 2018-02-21
Revert "New icon: Shutter"

jancavan:
Reverts Automattic/gridicons#279


jancavan:
The changes had to be reverted because `grunt` needs to be run.

PR #282, opened 2018-02-22
Bump to v2.1.3

jancavan:
Bumping Gridicons version to update Calypso.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "gridicons",
-  "version": "2.1.3-alpha.1",
+  "version": "2.1.3",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "gridicons",
-  "version": "2.1.2",
+  "version": "2.1.3-alpha.1",
```
gwwar: 2.1.3

gwwar:
Good to merge after the package.json is updated.

jancavan:
Thanks @gwwar! I actually updated it, but forgot to push my commit. Good catch!

PR #283, opened 2018-03-05
Allow Gridicons library to be imported by component

nosolosw:
The goal of this PR is to prepare Gridicons to be imported by component.

To use the external Gridicon at the moment we do: 

    import Gridicon from 'gridicons';
     <Gridicon icon="external" />

After this PR lands, with the new API we could also import the individual component directly:

    import GridiconExternal from 'gridicons/dist/external';
     <GridiconExternal />

By doing so, we're able to ship only the SVGs that the user needs and the size impact of using the Gridicons library in any project would become as minimal as it can be. For example, at the moment, to use the external Gridicon (278 bytes) in any project you'll pull 81K because the `Gridicon` component contains the whole icon set. By importing the individual component, you'll only pull 1,2K.

Although this introduces the ability to import Gridicons individually, it doesn't remove the current way of importing the whole iconset at once.

# How to test

* Delete your `build/` directory, if necessary.
* Run `npm run build`.
  * Verify that a `build/` directory containing untranspiled files exists.
  * Verify that a `dist/` directory containing files transpiled to ES5 and CommonJS module format exists.
* Run `npm pack` and verify that the `gridicons-2.1.3.tgz` contains: 
  * `dist` dir with the untranspiled files
  * `package.json`
  * `CHANGELOG.md`
  * `LICENSE.md`
  * `README.md`


jasmussen:
Really interesting, thank you for doing this. Question: can you still import _all_ gridicons at once, so we have backwards compatibility for existing code?

nosolosw:
Yes!

nosolosw:
Just a note for reviewers: this PR probably merits a new release. I can do that after review.

.npmignore
```diff
@@ -0,0 +1,12 @@
...
+docs/
+grunt-tasks/
+pdf/
+php/
+sources/
+react-icons/
+svg-min/
+svg-min-react/
+svg-sprite/
+Gruntfile.js
```
ockham: Is a blacklist the right approach here? Seems like we only want to publish only comparatively few files in the npm, which to me indicates a whitelist (i.e. the `files` attribute in `package.json` is better suited). Curious why you decided against continuing to use that?

package.json
```diff
@@ -4,7 +4,7 @@
   "main": "build/index.js",
   "scripts": {
     "build": "grunt --verbose",
-    "prepublish": "npm run build"
+    "prepublish": "npm run build && cp react-icons/*.js ."
```
ockham: Hmm, are we dropping all individual icon components into the package's root dir prior to publishing? That seems a bit unfortunate, especially if we don't have a cleanup job -- they're probably going to linger afterwards, requiring the dev to remove them manually...

Can we continue to keep things contained in dedicated build directories?

ockham:
How can we test this? I tried `npm run build` but ended up with a `build/` dir that didn't contain individual component files.

ockham:
Oh weird, on second try I actually got `react-icons/` instead of `build/` 🤔

sources/react/index-header-individual-component.jsx
```diff
@@ -0,0 +1,129 @@
...
+			'gridicons-stats',
+			'gridicons-status',
+			'gridicons-thumbs-up',
+			'gridicons-textcolor',
+			'gridicons-time',
+			'gridicons-trophy',
+			'gridicons-user-circle',
+			'gridicons-reader-follow',
+			'gridicons-reader-following'
+		];
```
ockham: I think we can shrink individual `*.jsx` file sizes even further if we don't include all the string arrays found here but rather move the logic found here to `grunt-tasks/svg-to-react.js`, which will allow us to only include the readily computed class names in the output files.

ockham:
Maybe we can also re-implement the generation of `react-icons/index.jsx` by simply ditching the `switch` / `case` statement there, and simply `import` all individual `*.jsx` files instead.

Gruntfile.js
```diff
@@ -143,7 +147,7 @@ module.exports = function( grunt ) {
           cwd: 'svg-min-react/',
           src: [ '**/*.svg' ],
           filter: 'isFile',
-          dest: 'build/'
+          dest: 'react-icons/'
```
ockham: Nitpick: We don't suffix our other build directories with `-icons` (since they all are icons, by definition 😉). So let's simply call it `react` or `jsx`.

sources/react/index-header-individual-component.jsx
ockham: What's even worse is that these lists are duplicated from `./index.jsx`. Redundancy === maintenance nightmare for everybody who's going to add a new icon. So we should really centralize this information and build individual files statically.

.npmignore
nosolosw: I've tried whitelisting every individual icon by setting the `files` prop to `*.js` and then blacklist only the `Gruntfile.js` using the `.npmignore`. It seems that in case of conflict the files prop takes precedence over .npmignore, so the Gruntfile.js was still included in the npm package. So I changed it to blacklisting to prevent the Gruntfile.js from being published.

package.json
nosolosw: Yes, they are moved to the root of the package only before `npm publish` (so it doesn't pollute it during development). We need that to do:

    import GridiconExternal from 'gridicons/external';

If they are under build, we'll need to include the `build` dir in the npm package and the individual icons would be imported like:

    import GridiconExternal from 'gridicons/build/external';

AFAIK, it is common practice for libraries that want to export individual components -such as Lodash- to move them to the root for publishing.

One thing we could do, though, is to run a clean task in the `postpublish` npm hook.

ockham: A `postpublish` cleanup hook is somewhat better, but still not perfect 😁-- as e.g. evidenced by [this thread](https://github.com/Automattic/gridicons/pull/283#discussion_r172834729).

Too bad `package.json`'s `main` field can't be a directory. Related: https://github.com/nodejs/node/issues/14970

Another idea: Have `build/index.js` do

```js
export { default as GridiconExternal } from './external';
export { default as GridiconFoo } from './foo';
export { default as GridiconGarb } from './garb';
```

If we publish not only the transpiled but also the untranspiled code to npm, consumers can then

```js
import { GridiconExternal, GridiconGarb } from 'gridicons';
```

For projects that use webpack >= v3, tree shaking will then ensure they're only importing the relevant code (omitting the Gridicons that aren't `import`ed).

.npmignore
ockham: (Related thread: https://github.com/Automattic/gridicons/pull/283#discussion_r172838047)

sources/react/index-header-individual-component.jsx
nosolosw: Good suggestion!

I'm glad that you mention the "adding new icon" task as the driving use case for this. At the time it wasn't evident to me how to centralize all that logic without introducing a lot of complexity for that particular use case - and that's the reason I stopped to hear my colleagues' opinions first. I think I may have an idea to solve that.

package.json
nosolosw: You raised a very good point about `cjs` and `esm` with https://github.com/Automattic/gridicons/pull/284 My response to that is a9010c937f93b56596c99721b2a78f5fefa967db :)

I still prefer to publish the icons separately, as not all consumers are using a module bundler that supports tree-shaking at the moment, and I'd want to impose that on them.

package.json
```diff
@@ -1,7 +1,10 @@
 {
   "name": "gridicons",
   "version": "2.1.3",
-  "main": "build/index.js",
+  "main": "cjs/index.js",
+	"files": [
+		"cjs"
+	],
```
ockham: Indentation 😁
I'd include a slash (`cjs/`) to make clear it's a directory (not just a file).
Can we also include `esm/`?

.gitignore
```diff
@@ -2,5 +2,5 @@ node_modules/
 /npm-debug.log
 .idea/
 svg-min-react/
-build/
+react-icons/
```
ockham: Needs to be
```
cjs/
esm/
```

instead of `react-icons` now.

ockham:
Thanks a lot for adopting the `cjs/` and `esm/` idea!

nosolosw:
@ockham This is ready for the second round, all your suggestions have been implemented - thanks for taking the time! I'll prepare a separate PR for bumping the version number once this has landed.

sources/react/template-devdocs-example.js
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * External dependencies
+ */
+import React from 'react';
+
+/**
+ * Internal dependencies
+ */
+import Gridicon from './index.js';
```
ockham: Can we actually just drop the `.js` extension?

gwwar:
> I'll prepare a separate PR for bumping the version number

I think it's fine to bump the package.json to an alpha in this PR, and if we publish with an alpha tag it would let us test this against a WIP Calypso PR. From the changes here, I think a major version in increase makes sense eg:

```
{
   "name": "gridicons",
   "version": "3.0.0-alpha",
```

sources/react/template-devdocs-example.js
nosolosw: By leaving the extension we make sure it works in any native environment (and doesn't depend on a bundler's ability to work without the extension, etc).

nosolosw:
@gwwar @ockham published it to npm and companion Calypso PR ready for review at https://github.com/Automattic/wp-calypso/pull/23171

sources/react/template-devdocs-example.js
ockham: Hmm. Sure about this? Wouldn't any native environment that supports ES Modules also be able to resolve this without the extension? After all, there are a number of extension-free `import`s, aren't there?

nosolosw: What I [read](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import) is that browsers won't understand that without the extension.

nosolosw: I'm happy to change it if it makes things clear, though.

README.md
```diff
@@ -24,17 +26,29 @@ render() {
...
 ```
 
+Or import icons individually:
+
+```
+import GridiconAddImage from 'gridicons/cjs/add-image';
+//...
+render() {
+    return <GridiconAddImage />;
+}
```
ockham: I think you can drop the `render()` {}` example here, it doesn't really make much sense outside the context of a component. I think it's fine to simply show what the `import` will look like here.

grunt-tasks/svg-to-react.js
```diff
@@ -3,52 +3,140 @@
...
+		const iconsThatNeedOffsetX = [
+			'gridicons-arrow-down',
+			'gridicons-arrow-up',
+			'gridicons-comment',
+			'gridicons-clear-formatting',
+			'gridicons-flag',
+			'gridicons-menu',
+			'gridicons-reader',
+			'gridicons-strikethrough'
+		];
```
ockham: Can we move the arrays to a dedicated file that we import here?

sources/react/template-all-icons.js
```diff
@@ -0,0 +1,55 @@
+const toString = array => array.reduce( ( acc, item ) => ( acc = acc + "'" + item + "', " ), '' );
+
+const prepareAllIcons = ( {
+	components,
+	iconsThatNeedOffset,
+	iconsThatNeedOffsetX,
+	iconsThatNeedOffsetY,
+} ) => `
+/** @ssr-ready **/
```
ockham: The `/** @ssr-ready **/` pragma is obsolete -- let's remove it while we're here (and add a note to the `Changelog` 😄)

sources/react/template-individual-icon.js
```diff
@@ -0,0 +1,48 @@
...
+const needsOffset = ( name, icons ) => icons.indexOf( 'gridicons-' + name ) >= 0;
+
+const prepareIndividualIcon = ( {
+	name,
+	component,
+	iconsThatNeedOffset,
+	iconsThatNeedOffsetX,
+	iconsThatNeedOffsetY,
+} ) => `
+/* @ssr-ready */
```
ockham: The `/** @ssr-ready **/` pragma is obsolete -- let's remove it while we're here (and add a note to the `Changelog` 😄)

sources/react/template-all-icons.js
```diff
@@ -0,0 +1,55 @@
...
+
+		${ components }
+
+	}
+
+	return ( svg );
+}
+`;
+
+module.exports = prepareAllIcons;
```
ockham: What about changing the this template to `import` all individual icon files, and then something like

```jsx
import React from 'react';

${ imports }

export default function( { icon: iconProp, className, onClick, size = 24, ...otherProps } ) {
	const IndividualGridicon = toCamelCase( icon ); 
	return <IndividualGridicon className={ className } onClick={ onClick } size={ size } ...otherProps />;
}
```

(You can then also omit `icon` from the destructuring in the individual templace.)

README.md
nosolosw: I prefer to keep it: it instructs the consumer of the API how to use the component noting that in the general case you'll use the icon prop and if you import the individual component you don't. That's pretty clear to my eyes, but I'd rather err on the side of being obvious for new users.

sources/react/template-all-icons.js
nosolosw: I haven't done it because my understanding is that the CommonJS `index.js` will be bigger: you are embedding react components that contain the svg icon. This way, we directly insert the svg icon.

grunt-tasks/svg-to-react.js
nosolosw: Done at bed88b4

sources/react/template-individual-icon.js
nosolosw: Done at d5963b9

sources/react/template-all-icons.js
nosolosw: Done at d5963b9

sources/react/template-devdocs-example.js
nosolosw: Done at d2d1478

sources/react/template-all-icons.js
ockham: It'd just be nice to get `import { GridiconFoo } from 'gridicons';`. But yeah, let's leave it as-is for now; we can always iterate.

ockham:
This is looking awesome! Great changes -- I love that we now have template files instead of header and footer ones.

One note, I'm seeing this pattern a lot in the generated `esm/`:

```js
	const iconClass = [
		'gridicon',
		'gridicons-align-center',
		className,
		false ,
		false,
		isModulo18( size ) ? 'needs-offset-y' : false,
	].filter( Boolean ).join( ' ' );
``` 

Think it's worth filtering the hard-coded `false` values at `svg-to-react` stage? Otherwise, I'd say this is ready to 🚢!

nosolosw:
Ah, that's something I've noticed as well. Tried to change it but made the code less readable (I like that the template is explicit about when an icon needs offset).

Will leave this open until https://github.com/Automattic/wp-calypso/pull/23171 is approved as well. Then, will change the gridicon package version in both before merging.

README.md
ockham: What about making it a functional component to have a more viable/standalone example? 😁

ockham: E.g.

```js
import GridiconAddImage from 'gridicons/cjs/add-image';
//...
function MyComponent() {
    return <GridiconAddImage />;
}

```

nosolosw: I've had too much npm for today :sweat_smile:

gwwar: I think the render() is more accessible for folks who have just started with React, let's keep in it. :)

gwwar:
Fantastic work @nosolosw! Thanks so much for taking this on. ❤️ 

I did some smoke testing in Automattic/wp-calypso#23171 and things look pretty solid. Let's wait for a final sign off from flow (they're great at breaking stuff ✨), and try to land this next early next week.

As an aside @folletto @jasmussen, do you think we could eventually move over social icons into this repo now that we can import a single icon at a time? social-logos has a very similar build and tends to lag behind gridicons.

folletto:
> As an aside @folletto @jasmussen, do you think we could eventually move over social icons into this repo now that we can import a single icon at a time? social-logos has a very similar build and tends to lag behind gridicons.

I'm not sure I understand why: the reason they are split isn't technical. These icon sets are based on entirely different principles and especially logos are regulated by specific copyright laws. 

I wouldn't merge the two unless there's a very compelling reason and a solution that covers all the concerns that separated them.

I would work in making the process of aligning the build systems easier — as we did on so far.

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 _The versioning refers to the React component build._
 
+#### v3.0.0 (2018-03-09)
+* React: remove obsolete @ssr-ready pragma from components.
+* React: allow importing individual icons, both in CommonJS and ECMAScript module formats.
+* React: substitute React.PureComponent for a JavaScript function, to minimize the bundle size of transpiled CommonJS files as much as possible.
+* Package: added a `esnext` key in the package.json so API consumers can use it for importing the main ESM file.
+* Build: use template literals to create the React components and centralize the _icon needs offset_ logic in the svg-to-react Grunt task.
+* Build: deleted the `build/` directory in favor of separate `cjs/` and `esm/` folders that contain CommonJS and ECMAScript modules respectively.
+
```
folletto: Can we avoid acronyms (cjs, esm) in favor of a clear, spelled out folder name?

* `common-js`
* `ecmascript`

Also: can we add somewhere a detail on why there's this split, and as a developer when should I pick one or the other?

In genereal: is there any way to avoid this split for a simpler solution that reduces a choice requirement?

nosolosw: That's a very good question I myself pondered about. 

We can have a default `import GridiconExternal from 'gridicons/external'` that pulls the CommonJS module (which is now the common format used in npm packages). That was actually the original proposal, but it would compromise future backward compatibility, as it was raised by @ockham Some notes on this:

* We live in times of change when it comes to JavaScript module formats. Many libraries are shipping both `cjs` and `esm`. There is no a strong standard on how to do it yet. I wish we could just avoid this module war, but we can't.
* If we publish now the `cjs` icons under `gridicons`, when `esm` becomes more used we won't be able to use that namespace and maintain backward compatibility. We'll need to publish the individual icons under `gridicons/esm`, for example. This would put us in a weird position and could potentially introduce errors if the consumer is expecting a `esm` module when using `gridicons/external`, for example.
* In the README, the `esm` option is deemphasized, which was intentional to mitigate decision fatigue.
* Consumers of the API can always teach their module bundlers to make `gridicons` an alias for `gridicons/cjs` or `gridicons/esm` - this is probably what we'll do in Calypso.

Does this address your questions @folletto ?

ockham: Yep, this is totally on me @folletto 😄

@nosolosw summed it up excellently. I'll just add that having untranspiled code available in `esm/` allows the consuming project to make use of features such as tree-shaking. I.e. with some further iterations, we'll be able to do handy named imports such as

```js
import { GridiconExternal, GridiconGarb } from 'gridicons';
```

while still ensuring that only the code that's relevant for these imports is being pulled in, thanks to tree-shaking -- i.e. we can get both nice named imports, _and_ smaller bundle sizes. This is something we can't do with CommonJS.

(For reference, my original suggestion was at https://github.com/Automattic/gridicons/pull/283#discussion_r172894269)

The different (CJS vs ESM) entrypoints are exposed through the `main` and `esnext` fields in `package.json`; the general idea is taken from http://2ality.com/2017/06/pkg-esnext.html (and as Andrés said, we can use e.g. webpack aliases to still allow 'pretty' import directory filenames, e.g. without the explicit `cjs/` or `esm/` parts).

I've also seen _some_ (albeit not a lot) of standardization towards these names ( `cjs/` or `esm/`), so I'd rather keep them. I'd like to think of them as an implementation detail that typically is aliased by the consumer anyway.

folletto: Thank you for the reply. The technical answer is grounding for the question I have. I'm concerned on how cryptic now this is. Before "react/build" was clear. Now there's choice, with two three-letter shorthands that require knowing _all of what you described_ to make a choice.

Granted, it's going to be packaged and published on npm, but is there any way to make this more understandable also in this library?

Also: I'm not sure if this is the case but when I read "it would compromise future backward compatibility" are we talking about Calypso or...? If Calypso already works with the more modern format, I'd suggest to ship just that and bump the requirements if needed. Or, at least, create a strong default for that, and leave the other as a secondary option.

At the very very least: can we pick a default, and have a one line explanation on when to use each of these two? Imagine a developer coming in and having no idea on the difference between the two when you write these one liners.

ockham: 'Default' now depends on context: If you project is CommonJS/ES5 only, use `cjs/`. If it supports ES6 stuff -- most notably modules -- either natively, or thru transpilation, use `esm/`. This is somwhat exposed by the relevant fields in `package.json`, but I guess we can also document in the `README`. Sounds good?

folletto: ...which one is Calypso?

Btw, how does that distinction work for people downloading from npm?

ockham: Well Calypso is transpiled using Babel, so it would be `esm/`. (It'll still work with `cjs/`, which is what our intended upgrade path will be like, to minimize side effects: Merge this, publish to npm, bump Calypso's dependency, and only then switch to `esm/`.)

> Btw, how does that distinction work for people downloading from npm?

Not sure I'm totally getting this. To make things a bit clearer, you could view `esm/` and `cjs/` as basically holding the results from different stages of the build process: `react-to-svg` results in ES6+ files which go to `esm/` (and can be readily used by projects that support those). These files are then transpiled using Babel, and the results end up in `cjs/`. Previously, both types of files would'be ended up in `build/`, but only the transpiled ones would be published to npm. We're essentially now just also including the untranspiled ones. The way we're doing this -- using separate directories and entry points -- mostly reflects constraints from today's JS, and the transitional period we're finding ourselves in, where we cannot yet fully remove transpiled files from an npm, since we might break some consumers which rely on code to be ES5/CommonJS.

nosolosw: To sum up, from the consumer point of view we have two alternatives.

The current proposal:

    // live as it was: import the whole iconset in CommonJS format
    import Gridicon from 'gridicons'; 
     <Gridicon icon="external" />

    // Import individual icons in CommonJS format.
    // This is what the README says you should do, 
    // so it'll end up being the option chosen by default.
    import GridiconExternal from 'gridicons/cjs/external';
     <GridiconExternal />

    // Import individual icons in ECMAScript format
    import GridiconExternal from 'gridicons/esm/external';
     <GridiconExternal />

An alternative approach:

    // live as it was: import the whole iconset in CommonJS format
    import Gridicon from 'gridicons'; 
     <Gridicon icon="external" />

    // A nicer path to import CommonJS which is the expected format atm
    import GridiconExternal from 'gridicons/external';
     <GridiconExternal />

    // ECMAScript individual icons
    import GridiconExternal from 'gridicons/esm/external';
     <GridiconExternal />

In addition to what was already said, I'll add that:

* with the alternative, every icon needs to be in the root of the repository. So, after `npm run build` we end up with dozens of files there - to clean that up you need to run a command. This is not the "external API" of the library but it also worth considering the developer experience for Gridicons when talking about this. See this [convo](https://github.com/Automattic/gridicons/pull/283#discussion_r172838047).
* whoever downloads Gridicons with npm to use the React components in their project needs to prepare that code for browsers. It's impossible to use this code without a transpiling tool. Whoever sets up the config, already knows about and needs to be aware of CommonJS / ECMAScript formats.

As I see it, we have a trade-off to make. We either accept breaking backward compatibility in the future when this _module transition_ is over (and go with the alternative now and switch the paths in the future), or we are agnostic about modules (the current proposal). Taking into account that users of the React components already have that knowledge (or copy-paste the README examples, so they're fine) I lean towards the current proposal.

@folletto does it make more sense? would you still prefer the alternative to the current approach?

jsnajdr: There are several ways how to export stuff from a NPM module:

1. Exporting a CommonJS module: that means using `export.something =` and `module.exports =` statements to export and using `require()` to import dependencies. Code is usually transpiled to ES5. Such a module can be consumed by Node.js directly, and also by all browser bundlers. The oldest and most common format. In `package.json`, the `main` field specifies the location of the module file for this format.

2. Exporting ES6 modules. Code is **still transpiled to ES5**, but imports and exports are done by the respective ES6 statements. These are the only bits of ES6 in an otherwise ES5 code. This module cannot be consumed by Node, and is intended for browser bundlers like Webpack. The bundler will find all the imports and exports, resolves them and outputs an ES5 bundle. No transpilation needs to happen here. This format gradually replaces Common JS for browser libraries. The advantage is that bundlers can optimize ES6 modules much better (tree shaking, marking exports as pure...). In `package.json`, the module file is specified by the `module` field.

3. Exporting untranspiled ES6 modules. Recent versions of Node.js and of major browsers (Chrome, Edge, Firefox, Safari) have full ES6 support, so transpiling is not needed and providing an ES6 version of your module makes sense. That's the `es:next` field and it's very experimental -- there is little mainsteam usage at this moment.

The distinction between option 1 and option 2 is important -- there's been some misconception in this convo that mixes them together.

Gridicons don't need to specify an `es:next` field. What we really want here is `module`.

jsnajdr: How should the import syntax for Gridicons look like? I believe that the `cjs` and `esm` strings shouldn't be a part of the module's user interface at all. That's an internal detail of the module that Webpack negotiates with its `package.json` and chooses the right option.

Does the ideal interface look like this?
```js
import Gridicon from 'gridicons'; // full package, default import
import { GridiconExternal } from 'gridicons'; // named import
// alternative, less that ideal IMHO
import GridiconExternal from 'gridicons/external'; // import from subpackage
```
or in CommonJS version like this?
```js
var Gridicon = require( 'gridicons' );
var GridiconExternal = require( 'gridicons/external' );
```

Given that we want the bundle to be optimized, i.e., unused icons are not bundled:
- the ES6 ideal above, with one `gridicons` package, is implementable only in Webpack 4 with its `sideEffects` flag.
- the subpackage ES6 import can be implemented without any trouble
- the CommonJS version with one `gridicons` package cannot be efficiently implemented at all
- the CommonJS version with subpackages can be implemented only by having built files in the root directory, with no `build` or `dist` module.

Is it an option to provide the "import single component" feature only for ES6 module consumers? And the CommonJS consumers always getting the whole package? That would simplify things a lot.

nosolosw: @jsnajdr thanks for taking the time. I'm trying to distill your comment to some actionable items. Let me know if this makes sense.

> Is it an option to provide the "import single component" feature only for ES6 module consumers? And the CommonJS consumers always getting the whole package?

The user-facing need I'm trying to address is to help Gridicons to ship only the code that the user needs. I also want to make this work for every project that uses Gridicons, independently on the state of any specific bundler/tooling/project/etc.

With that in mind: 

* I consider exporting individual icons in CommonJS format a core part of the proposal, because that's the ubiquitous format at the moment that most tools understand and are capable of working with.
* I've left out named ES6 imports because they don't provide size reduction unless you use a specific set of tools and config. I see that as a nice-to-have. To control scope and I'd like that to be out of this PR.
* I'm open to remove the `esnext` key in the `package.json` if that's a concern. `esm` modules would be still importable without it and I agree that's experimental and not mainstream yet. It makes sense introduce it later when/if the ecosystem has made its mind.
* Although I consider the current proposal more future-proof and with a better developer experience, I won't be opposed to the alternative [I've outlined here](https://github.com/Automattic/gridicons/pull/283#discussion_r173806612). I'm happy to discuss any other option that accomplishes the same goals.

So, @jsnajdr would you agree to land this PR as it is but removing the `esnext` key in the `package.json`?

folletto: I feel this conversation expanded a little bit — which is great to explore all the details, but I want to make sure we are all on the same page on the why I started this thread: 

1. We are adding a performance feature to import single icons — that's awesome.
2. Yet, this is a icon set. We are introducing two new folder structures, and a new syntax. Are we sure all this complexity is needed for an icon set? (probably yes) We exploded from a single nice `build` folder with react data, to... well all the complexity of the lengthy discussion above.

To be clear: I'm **not** dismissive here – I know that while I can understand the points above I can't argue pros and cons. I just want to stress that we are aiming for a simple approach, intuitive to both maintain and use, and I need your help to review if a simpler approach is possible, or not. I'm totally fine if the answer is "no", I just want to avoid for the answer to be "no" by default just because we haven't tried. :)

ockham: > `import { GridiconExternal } from 'gridicons'; // named import`

This is definitely my preferred style.

> the ES6 ideal above, with one `gridicons` package, is implementable only in Webpack 4 with its `sideEffects` flag.

Bummer. I honestly wasn't aware of that 🙁  I thought tree-shaking would work in this case with Webpack >= 3 (2?).
(I might not seeing the whole picture, but doesn't the same hold true for any kind of import? In what cases is it required to specify the `sideEffects` flag to actually get tree-shaking?)

folletto: I clarified a few extra points with Bernie in DM as I know he understand my line of questioning that might have ended up a bit too heavy for this thread. 

So here's an idea to move this forward in a way that should balance all the requirements:

1. Keep `cjs` and `esm` folder structure (thanks for the clarifications on this!)
2. Update the PR to support the preferred syntax (the one in the comment above by Bernie)
3. Update the readme to make the preferred syntax explicit.
4. Update the readme with a one-line explanation of what `cjs` and `esm` folders are for, and how are they used (which includes a note on the alternative, but unsupported, syntax).

ockham: > `import GridiconExternal from 'gridicons/external'; // import from subpackage`
>
> the subpackage ES6 import can be implemented without any trouble

@jsnajdr But wouldn't that require moving the individual ESM files to the package root (or an alias on the consumer side) as well?

nosolosw: For me, consumers are any project that wants Gridicons, not only Calypso. I actually started this out of a particular need for [happychat-client](https://github.com/Automattic/happychat-client/issues/10) so doing this wouldn't be worth for me it I can't use it there.

I'm committed to make this PR work because it addresses a real need. So, here's one bold thing we could do: let's forget completely about the ECMAScript modules. This is precisely what we are doing in Gridicons at the moment and perhaps it's unnecessary to introduce that complexity right now. The [two options above](https://github.com/Automattic/gridicons/pull/283#discussion_r173806612) would become.

- If we use a folder to contain the icons so the devexp is nicer and we have future backward compatibility with `esm` (I use `dist/` instead of `build/` because is more common):

      // Live as it was: import the whole iconset
      import Gridicon from 'gridicons'; 
      <Gridicon icon="external" />

      // Import individual icons
      import GridiconExternal from 'gridicons/dist/external';
      <GridiconExternal />

- If we publish the icons in the root of the repository (nicer paths but will break backward compatibility when we switch Gridicons to ECMAScript modules):  

      // Live as it was: import the whole iconset
      import Gridicon from 'gridicons'; 
      <Gridicon icon="external" />

      // Import individual icons
      import GridiconExternal from 'gridicons/external';
      <GridiconExternal />

To be honest, I'm not the most qualified person to choose between the two because it boils down to how we want to manage the Gridicons library: do we want backward compatibility or not? Being this PR my first contribution I assumed that we wanted that, but it's totally fine if we don't.

@folletto would this simplify things?
Any of the devs consider this as a no-go? @ockham @jsnajdr @gwwar

nosolosw: @folletto @ockham wasn't aware of your comments when I posted [this](https://github.com/Automattic/gridicons/pull/283#discussion_r174093348). Does it modifies your thinking @folletto or you still hold on [this](https://github.com/Automattic/gridicons/pull/283#discussion_r174088976)?

jsnajdr: > But wouldn't that require moving the individual ESM files to the package root (or an alias on the consumer side) as well?
@ockham Good point, that was a mistake on my side. Both CJS and ES subpackages require moving files to package root. I edited my comment accordingly.

ockham: @nosolosw WFM!

jsnajdr: > In what cases is it required to specify the sideEffects flag to actually get tree-shaking?)

In case there is a top-level `index.js` that reexports submodules:
```js
export { default as GridiconNotice } from './notice';
export { default as GridiconSearch } from './search';
```
and we import only `GridiconNotice`, then how does Webpack and UglifyJS know they can omit the `GridiconSearch` import? They don't, because importing the `./search` module can have side effects. For example, it can change some global variable or register a DOM listener.

The `sideEffect: false` flag says that no such side effects exists and unused imports can be safely removed.

nosolosw:
Summing up several threads and private convos with @folletto , @ockham and @jsnajdr this is the agreement:

* We're going to forget about `esm` modules in this PR, but want any code to be API backward compatible in the future.

* This is the ideal API we're aiming for:

      // the whole iconset
      import Gridicon from 'gridicons';
      <Gridicon icon="external" />

      // a particular icon
      import { GridiconExternal } from 'gridicons';
      <GridiconExternal />
   
The problem is that it doesn't allow us to reduce size with the majority of tooling, which was the goal of this PR. So we'll provide an escape hatch which is backward compatible: 

    // a particular icon
    import GridiconExternal from 'gridicons/dist/external';
    <GridiconExternal />

I'll look into adding `import { GridiconExternal } from 'gridicons';` to this PR. If it's too much work that will be left out for a later PR and we land this anyway, so we can start having the benefits of size reduction.

CHANGELOG.md
jsnajdr: @nosolosw Exporting just the CJS modules and forgetting about ES6 now works for me. The syntax would be:
```js
import Gridicon from 'gridicons';
import GridiconExternal from 'gridicons/dist/external';
```

In future, when Webpack 4 becomes ubiquitous, we can migrate to full ES6:
```js
import { GridiconExternal } from 'gridicons';
```

In the meantime, we could have a Babel transform that lets us write the named import syntax right now and transform it to a `gridicons/dist/${icon}` import. We have something like this for Lodash and Redux Form already in Calypso. But that's optional.

I hope I summarized everything correctly :)

gwwar:
Let us know when you're ready for another look!

Gruntfile.js
```diff
@@ -143,7 +147,7 @@ module.exports = function( grunt ) {
           cwd: 'svg-min-react/',
           src: [ '**/*.svg' ],
           filter: 'isFile',
-          dest: 'build/'
+          dest: 'dist/'
```
ockham: I'm fine with moving the (CJS) stuff we're going to publish to `dist/`; however, what about still building the intermediary ES6 files (that, IIUC, we aren't publishing for now) in `build/`? Just to keep things a bit more separate -- the `dist/` dir will then only hold stuff that we're also going to publish. That will allow us to set `"files": [ "dist/" ]` instead of `"files": [ "dist/*js" ]` in `package.json`. (IMO, the latter is both uglier, and more fragile, since we're relying on `*.js` (CJS) vs `*.jsx` (ES6) extensions, which is quite subtle and easy to miss.)

nosolosw:
Released 3.0.0-alpha.2 to npm without the named imports part - that is left for the future. Companion Calypso PR https://github.com/Automattic/wp-calypso/pull/23171 was updated as well.

This is ready for review and, hopefully, land!

nosolosw:
@ockham Done at b884c31. Merged this PR and published to npm `3.0.0`.

Thanks everyone for participating and for welcoming a newbie to the gridicons repo! I know this has taken longer than expected, but the topic is complex and I appreciate the effort everyone has done to understand each other until we've built a shared context and consensus. :heart: 

Now, let's profit from this size reduction everywhere.

folletto:
Thank you for your openness to discussion and motivation to push this through, you're 🌟

gwwar:
Awesome work @nosolosw !

PR #285, opened 2018-03-12
Add Multiple Users Icon

SylvesterWilmott:
Adding an icon for multiple users

multiple-users-example


folletto:
Looks good.

Do a quick review of the SVG source, some details here: https://github.com/Automattic/gridicons/wiki/Guidelines-for-SVG-Sources

No biggie: looking at it seems just we should remove the `id=` and maybe the XML encoding declaration.

drw158:
Looking good!

SylvesterWilmott:
@folletto rectified those parts. Should be good now

folletto:
Ok let's build and merge ;)

gwwar:
Please keep an eye on large line diffs 😄 @folletto @SylvesterWilmott 

Looks like we added a package.lock here. Do we need it? This file does not get published with the npm package, but it is checked in locally. If we happened to have an unpinned library version, we wouldn't be able to see the errors locally if a bugfix version happened to break something, but it would show up in other projects that use gridicons as a dep.

folletto:
> Looks like we added a package.lock here. Do we need it? This file does not get published with the npm package, but it is checked in locally. If we happened to have an unpinned library version, we wouldn't be able to see the errors locally if a bugfix version happened to break something, but it would show up in other projects that use gridicons as a dep.

Can you open a separate issue? I'm not sure if we need it, and we already had issues about it: https://github.com/Automattic/social-logos/pull/58 but we haven't clarified if we need it at all.

folletto:
> Please keep an eye on large line diffs

What do you mean?

gwwar:
@folletto 😄 package.lock added ~5k here, it's fine if we want it in the project, but I noted that there was no discussion about it.

screen shot 2018-03-12 at 9 30 39 am

folletto:
Ah ok it was referring to the same thing. I didn't discuss it here because I thought we were already having that discussion in the other repo, and since we added it there, I thought it was fine aligning and adding it here. We need to have that discussion ;)

PR #289, opened 2018-03-14
Fix/main gridicon component

nosolosw:
By changing the component to be a functional one in https://github.com/Automattic/gridicons/pull/283, we broke backward compatibility. That resulted in a break in icons that used the React ref to keep track of the DOM element.

This PR reverts it to a React.PureComponent.

### How to test

https://github.com/Automattic/wp-calypso/pull/23306 makes use of `3.0.1-alpha.3` and is a good testbed.

### DIFF
--- a/sources/react/template-all-icons.js
+++ b/sources/react/template-all-icons.js
@@ -1,52 +1,67 @@
-const toString = array => array.reduce( ( acc, item ) => ( acc = acc + "'" + item + "', " ), '' );
+const toString = array =>
+        array.reduce((acc, item) => (acc = acc + "'" + item + "', "), '');
 
-const prepareAllIcons = ( {
-	components,
-	iconsThatNeedOffset,
-	iconsThatNeedOffsetX,
-	iconsThatNeedOffsetY,
-} ) => `
+const prepareAllIcons = ({
+        components,
+        iconsThatNeedOffset,
+        iconsThatNeedOffsetX,
+        iconsThatNeedOffsetY,
+}) => `
 /**
  * External dependencies
  */
-import React from 'react';
+import React, { PureComponent } from 'react';
+import PropTypes from 'prop-types';
 
-export default function( { icon: iconProp, className, onClick, size = 24, ...otherProps } ) {
+export default class Gridicon extends PureComponent {
 
-	const iconsThatNeedOffset = [ ${ toString( iconsThatNeedOffset ) } ];
+	static defaultProps = {
+		size: 24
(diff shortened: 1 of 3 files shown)


sources/react/template-all-icons.js
```diff
@@ -1,52 +1,69 @@
...
 
-	const doesItNeedOffset = ( name, icons ) => icons.indexOf( name ) >= 0;
-	const isModulo18 = size => size % 18 === 0;
+		const { size, onClick, icon: iconProp, className, ...otherProps } = this.props;
 
-	const icon = 'gridicons-' + iconProp;
-	const needsOffset = doesItNeedOffset( icon, iconsThatNeedOffset ) && isModulo18( size );
-	const needsOffsetX = doesItNeedOffset( icon, iconsThatNeedOffsetX ) && isModulo18( size );
-	const needsOffsetY = doesItNeedOffset( icon, iconsThatNeedOffsetY ) && isModulo18( size );
+		const iconsThatNeedOffset = [ ${toString(iconsThatNeedOffset)} ];
```
blowery: might be better to have these as a static property of the class. That way you're not recreating the array on every render.

sources/react/template-all-icons.js
```diff
@@ -1,52 +1,69 @@
...
+		const iconsThatNeedOffsetX = [ ${toString(iconsThatNeedOffsetX)} ];
 
-	switch ( icon ) {
-		default:
-			svg = <svg height={ size } width={ size } { ...otherProps } />;
-			break;
+		const iconsThatNeedOffsetY = [ ${toString(iconsThatNeedOffsetY)} ];
 
-		${ components }
+		const doesItNeedOffset = ( name, icons ) => icons.indexOf( name ) >= 0;
```
blowery: these can both be module level functions since they don't use `this` or anything from the closure they create.

blowery:
Found a couple little optimizations in the proposed render method. They seem to have existed prior, so it's not really a priority to fix them here... But would we want to address it now, or in another PR?

sources/react/template-all-icons.js
nosolosw: Done at 9af3e7d8ada9e7877e8baaaecf4ebf8e7554b40b

sources/react/template-all-icons.js
nosolosw: Done at 9af3e7d8ada9e7877e8baaaecf4ebf8e7554b40b

PR #291, opened 2018-06-25
Add alternative stats icons

porada:
At the Studio, we use Gridicons when designing the WordPress.com landing pages.

Some of our pages repeatedly mention _growth_ or an _increase_ of something over time. `stats` and `stats-alt` often accompany those sections, but I don’t think they cut it. I’d like to propose two alternative states for those two icons to provide some more flexibility and to help us visualize the copy a little better.

Here’s the zoomed-in preview. The new icons are largely based on the originals.

Icon preview

Please me know what you think. ✌️


porada:
BTW, I’m not too happy about the naming convention for these alternatives, but I don’t have a better idea that wouldn’t renaming the originals either. If you think they should be renamed, let me know.

drw158:
Thanks @porada, I don't see any reason not to add these in. I think the naming is fine too.

Let someone know if you'd like to add these in yourself, or if you need some help.

porada:
@drw158, if you could review the commit I added and then merge, that would be great. Thanks. 🙂

porada:
I went ahead and merged it! 🖖

PR #295, opened 2018-08-14
Updated icon: Filter (align with Material)

folletto:
This is a proposal to update the "Filter" icon to match the corresponding icon in Material design ([filter_list](https://material.io/tools/icons/?style=baseline)).

Why? Both icons are quite obscure. Since the other icon is in Material however, it might work better by reinforcement all across the Android ecosystem.

screen shot 2018-08-14 at 13 12 53

screen shot 2018-08-14 at 13 12 40

(I'll cleanup the SVG if we decide to go ahead).


folletto:
Note: this merge has a massive file count (~200) because likely some mismatch in the build system (there is a similar ~200 files change in #293, which might be the cause of this, but it's hard to tell now).

I did rebuild this by clearing out entirely my `node_modules` folder and re-doing `npm install`, which _should_ mean it's now fixed... but it's hard to say. 

Please keep an eye on any future massive change before merging: ideally any change to icons should exclusively rebuild the files related to that icon (i.e. the icon itself and the sprites).

PR #297, opened 2018-09-05
Updated pin icon

SylvesterWilmott:
Pin icon fix based on feedback [here](https://github.com/Automattic/gridicons/pull/296#issuecomment-418483425)

Changes:
- Orientation of icon is now vertical
- Path nodes cleaned up
- Style of pin is now slightly different

pin_example


folletto:
SVG wise it's ok.

I'll leave to Dave for a review of the icon itself 👍

drw158:
Thanks for the update. The only thing I might do is use sharp corners for the bottom part, just to align it a bit more with the rest of the set. I'll leave that up to you.

screen shot 2018-09-06 at 9 19 40 am

mattmiklic:
This is purely subjective feedback but compared to #296 the vertically-oriented icon is harder to understand at a glance. By virtue of being at a 45º angle that pin fills more of the available space and looks more balanced compared to other Gridicons.

Less important but also noticed -- the blunt, rounded terminal of the pin looks weird. It's more angled and "sharp" in the original, which seems more appropriate for a pushpin.

I realize that there is a balance between consistency with our icon standards and what "looks" best so I don't feel too strongly about these opinions, but I think they're worth mentioning.

SylvesterWilmott:
>the blunt, rounded terminal of the pin looks weird. It's more angled and "sharp" in the original, which seems more appropriate for a pushpin.

While I was working on the icon at a magnified size the appearance of a sharp point is way nicer but at 24px the difference is negligible with the rounded version appearing more pixel sharp

>This is purely subjective feedback but compared to #296 the vertically-oriented icon is harder to understand at a glance. By virtue of being at a 45º angle that pin fills more of the available space and looks more balanced compared to other Gridicons.

I'm still on the fence about this one, I understand and agree with both angled and vertical. At least for now I will merge the vertical type due to the fact that it has a lot less sub pixels and this icon should work at smaller sizes than 24px on mobile.

SylvesterWilmott:
@folletto This should be good to

folletto:
The build looks good! Thanks! 👍

PR #298, opened 2018-09-10
docs(readme): add syntax highlighting to code examples

flootr:
• add syntax highlighting to code examples
• add `$` to npm cmd to indicate shell usage

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ $ npm install gridicons --save
 
 You can either import the whole iconset and decide at run-time which icon to use:
 
-```js
+```jsx
 import Gridicon from 'gridicons';
 //...
 render() {
@@ -28,7 +28,7 @@ render() {
 
 Or import icons individually:
 
-```js
+```jsx
 import GridiconAddImage from 'gridicons/dist/add-image';
 //...
 render() {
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -10,15 +10,15 @@ Gridicon renders a single svg icon based on an `icon` prop. It takes a size prop
...
+```sh
+$ npm install gridicons --save
 ```
 
 #### Usage
 
 You can either import the whole iconset and decide at run-time which icon to use:
 
-```
+```js
```
nosolosw: Shall we use `jsx` instead?

README.md
```diff
@@ -28,7 +28,7 @@ render() {
 
 Or import icons individually:
 
-```
+```js
```
nosolosw: Same here.

README.md
flootr: I didn't know that existed. TIL thanks! :)

README.md
```diff
@@ -10,15 +10,15 @@ Gridicon renders a single svg icon based on an `icon` prop. It takes a size prop
 
 There's a gallery with all the available icons in http://automattic.github.io/gridicons/.
 
-```
-npm install gridicons --save
+```sh
+$ npm install gridicons --save
```
nosolosw: Perhaps we should not include the `$` bash mark? That makes difficult to copy-paste and I guess that highlighting it is bash code would be enough?

flootr: I agree that makes copying harder. Let's don't include it. Syntax highlighting doesn't give us much highlighting for shell commands so what about including a small comment that this is intended to be executed in a shell?

PR #300, opened 2018-09-14
try: publish 'svg-min' folder

flootr:
It seems we have [use](https://github.com/Automattic/wp-calypso/pull/27158) [cases](https://github.com/Automattic/wp-calypso/pull/27186) for accessing raw svg files in Calypso instead of React components. Ideally I think they should be included in the `dist/` folder as well, let's iterate on that.


flootr:
closing for now

PR #302, opened 2019-04-09
New icon: Site

drw158:
This icon is a simpler version of the themes icon, but it represents a user's site instead of the theme itself.

Screen Shot 2019-04-09 at 4 08 22 PM
Screen Shot 2019-04-09 at 4 07 54 PM


jasmussen:
This seems good to me. 👍 👍

folletto:
There seem to be far too many PDF generated. It should be only one, the one you added. 
I wonder if they updated again the library and broke it again... but could you check if you have the latest version of everything?

PR #305, opened 2019-04-10
Fix up the new site icon

blowery:
Remove the leading rect that causes the icon to appear as a black box after minification

To test, `npx serve docs` and then load the example page and check out the site icon. It should appear properly, not as a black box.


folletto:
Before:
Screenshot 2019-04-11 at 11 30 25

After:
Screenshot 2019-04-11 at 11 30 20

Thanks for the patch! All good! :D

PR #307, opened 2019-04-18
Add spritemap and offsets to package.

sgomes:
In Calypso, this will allow us to transition away from the SVG-in-JS icons to a better approach, making use of the spritemap with SVG external content.

The rest of the package is kept as-is, to maintain compatibility.

@blowery I think this is a good approach, but I'd like your input. It should be ok to include these files in the package and depend on them explicitly in Calypso, right?

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,6 @@
   "main": "dist/index.js",
   "files": [
     "dist/",
-    "sources/react/icons-offset.js",
     "svg-sprite/gridicons.svg"
   ],
   "scripts": {
--- a/package.json
+++ b/package.json
@@ -1,9 +1,11 @@
 {
   "name": "gridicons",
-  "version": "3.2.0",
+  "version": "3.3.0",
   "main": "dist/index.js",
   "files": [
-    "dist/"
+    "dist/",
+    "sources/react/icons-offset.js",
+    "svg-sprite/gridicons.svg"
   ],
   "scripts": {
     "build": "grunt --verbose",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,8 @@
 _The versioning refers to the React component build._
+### v3.3.0 (2019-04-18)
+* Include SVG spritemap (svg-sprite/gridicons.svg) in gridicons npm package.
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -1,9 +1,11 @@
 {
   "name": "gridicons",
-  "version": "3.2.0",
+  "version": "3.3.0",
   "main": "dist/index.js",
   "files": [
-    "dist/"
+    "dist/",
+    "sources/react/icons-offset.js",
```
blowery: this one feels a little odd. I think ideally it would be part of `dist/index.js`, or at least live in `dist/`. 

Probably ok for now though... hmmm.

blowery: Yeah, I think copying it to `dist` would be enough for now. That's where the current docs tell folks to import individual bits from already.

sgomes: No problem! I copied it to `dist/cjs` to distinguish from the other files, which have already been run through a bundler. How does that sound?

sgomes:
@blowery Any chance of a final review on this one so that we can move forward with SVG external content gridicons in Calypso?

package.json
blowery: why cjs? that didn't exist before...

sgomes: The current situation is already a bit problematic, since we have three things under `dist`: `index.js`, with all the icons, `example.js`, with a pre-packaged example with a bunch (maybe all?) icons, and then the individual `*.js` icon files. There's no real distinction between them, and the WebPack rule in Calypso was generating duplicates because of it (now fixed by explicitly excluding `example.js` from the bundle).

I'm reluctant to add yet another special file that lives in the same directory as everything else. Adding a new directory and placing `icons-offset.js` in there seemed like a good option, but `cjs` may indeed be a misleading name, since the other files are exposed as CommonJS modules too, now that I look closer. Would e.g. `util` be better?

sgomes: Here's what we currently ship in `dist/`, for reference:
https://unpkg.com/gridicons@3.2.0/dist/

blowery: I think `dist/util` would be a good spot. Agreed that this package needs a rework.

sgomes: Done! 👍

PR #309, opened 2020-10-15
Add icon for the Stories feature

bjtitus:
New icon:

gridicon-story

Stories are a new post type shown in the mobile apps (they are implemented as as a Gutenberg block).


bjtitus:
@folletto @davewhitley I was hoping one of you could point me in the right direction on generating the assets for these. When I run the script, it produces mangled PDFs and makes modifications to the svg files in the `docs` and `sprites` folders. Perhaps one of the dependencies has changed?

I am running on Big Sur, so I created a Docker image using the Node base image but still see the same thing.

Maybe my [package-lock file](https://gist.github.com/bjtitus/0a6782d59c4e1dfd150a2852eb0a9896) has some clues.

folletto:
Heya! I can confirm that... something changed. I get also all the PDFs being now updated, which shouldn't be the case. I feel some package has changed its internals at some point and... it broke... again.

It seems that ALL PDFs are generated broken now — all file sizes are 437 bytes.

Also... we are requiring a _specific_ version of both "svg-to-pdfkit" and "pdfkit" so... I'm really wondering what broke given the version has to be exactly that one.

folletto:
I created https://github.com/Automattic/gridicons/pull/310 to look into that.

folletto:
Ok #310 helped me find out the problem, #311 should fix it.

Could you please @bjtitus review #311 so we can merge that? 
Once it's done, you should be able to add the icon as expected.

davewhitley:
Might be relevant to choosing a specific pdfkit version https://github.com/Automattic/gridicons/pull/304

bjtitus:
@folletto Thanks! This seems to have fixed the generated PDF and it's not touching the other files. 🎊

folletto:
Nice. I checked the commit, and it seems alright. 

I think we can merge — are you ok doing it? :)

bjtitus:
Yep! Thanks checking it over!

PR #311, opened 2020-10-27
Build: update PDF async 'finish', fix stats-alt-2 icon

folletto:
Following from #309 and #310 I found out that at some point PDFkit changed from `close` to `finish` its completion event. This meant that the correct creation of PDFs was unreliable and just a matter of chance for the stream to write on disk. The outcome was a build process for PDFs that unreliably either made them correctly, or blanked them all (all PDFs set at 437 bytes).

This PR:

* Changes `close` to `finish`, restoring the proper async callback.
* Fixes a stray commit of the `stats-alt-2` icon (10956c1c1ae89a8591282469b2f5c21c4ee729ea) that for some reason was added directly to `svg-min` without a PR. Now it's properly in `svg` and in the build process.

Testing:

1. Try to `grunt` build and make sure no PDF gets updated.
2. Review all the PDFs to make sure they are all good.
3. Check `stats-alt-2` in all versions to make sure it works.


jasmussen (approved):
Code looks good, and testing instructions reveal no changed PDFs:

Screenshot 2020-10-27 at 15 15 28

And PDF files also look good. Nice.

PR #319, opened 2021-08-31
Update dependencies

noahtallen:
Since this hasn't been updated in several years, there are a few security problems.
- I've updated the the grunt and babel dependencies
- Switched to use `@babel/preset-env` and `preset-react` (instead of specifying presets and plugins manually)
- Fixed a charset problem in the `demo/template.html` file.
- Pinned the node and npm versions with volta
- Added the package-lock file.

The second commit is all of the files which changed when running `npm run build` after the first commit. Since we are updating the tools which generate/clean up the SVG files from the source svg, every downstream file (such as pdf, php, final svg, and react) also gets updated.

I verified the SVG + PDF output is correct by visually reviewing that there is no difference before and after.
 
### Testing
- [x] Verify that the generated PDF files are correct
- [x] Verify that the generated svg files are correct
- [x] Verify that it works with Calypso.
  - I created a calypso branch with the gridicon dependency changed to `https://github.com/Automattic/gridicons#try/update-dependencies`
  - I started calypso.
  - I visited http://calypso.localhost:3000/devdocs/design/gridicon, to test the gridicon component live.
  - I inspected an icon in devtools
  - I looked at a portion of SVG known to be changed in this PR -- verifying that the updated SVG is being loaded.
  - I verified that all the icons look good.


sources/demo/template.html
```diff
@@ -237,9 +238,7 @@ <h4>Icon names</h4>
 <div class="icons">
 {{#each icons}}
   <div>
-    <svg width="24" height="24" class="gridicon {{name}}">
-    <use xlink:href="gridicons.svg#{{name}}" />
```
noahtallen, line 241: `xlink:href` was deprecated and is not supported in modern html. Additionally, plain `href` does not support importing from files any more. It appears that `<object data="` is the way to go, but hard to verify since the file wasn't working on trunk.

scinos (approved):

PR #321, opened 2022-09-14
Add Reader engagement icons

eoigal:
This PR aims to add new icons that will be used in the Reader as part of a redesign project.

### New icons

* Reader Comment
* Reader External
* Reader Share
* Reader Star

### Updated icons

* Reader Follow
* Reader Following


jasmussen:
Ran this through `npm run build`, and the build process appears to be working. Two things we need to do, however:
Screenshot 2022-09-15 at 18 03 21

All these files after the grunt build process needs to be committed as well.

However, it looks as if the source SVG files need to be edited, looks like they are including a big square surrounding the entire artboard that needs to be deleted. SVGO when it removes the color definitions means the icons become black squares:

Screenshot 2022-09-15 at 18 04 13

I also still think it's a little curious that both the new stars are outlined. Both in terms of their intended use — could a `stroke` be applied to the single star and save one of the icons from being necessary? There are also some stars already:
Screenshot 2022-09-15 at 18 20 55

Nothing I think needs to block this PR, but the value of the extra stars isn't entirely clear.

sources/svg/gridicons-reader-following.svg
```diff
@@ -2,10 +2,12 @@
...
-		c0,1.1,0.9,2,2,2h4.538l4.913-4.832L15.545,15.344z M8,13H4v-1h4V13z M11,11H4v-1h7V11z M11,9H4V8h7V9z M18,6H4V4h14V6z"/>
-</g>
-<g id="Layer_1">
-</g>
+	<g id="reader-following">
+		<mask id="mask0_372_400" style="mask-type:alpha" maskUnits="userSpaceOnUse" x="3" y="3" width="21" height="19">
+			<path fill-rule="evenodd" clip-rule="evenodd" d="M21 11.889L16.545 16.344L14.451 14.168L9.538 19H5C3.9 19 3 18.1 3 17V3H21V11.889ZM16.508 22L24 14.482L22.612 13.106L16.518 19.2L14.412 17.012L13 18.4L16.508 22ZM9 14H5V13H9V14ZM5 12H12V11H5V12ZM12 10H5V9H12V10ZM5 7H19V5H5V7Z" fill="white"/>
+		</mask>
+	</g>
+	<g mask="url(#mask0_372_400)">
```
anomiex, line 10: It seems some tooling knows what to do with the mask here while some doesn't, resulting in a black square in some cases rather than the intended image.

I simplified this in 59e169a to use the path directly as the image instead of masking a black square.

PR #324, opened 2023-03-15
Add a new science icon for staging sites

katinthehatsite:
I would like to add a new science icon to the Gridicons. This icon will be used for staging sites when they are launched:

Screenshot 2023-03-15 at 5 01 29 PM

* Issue: https://github.com/Automattic/dotcom-forge/issues/1962 -> the icon will be implemented in that issue
* Figma file: https://www.figma.com/file/6H9TZJNUl2sqTKO7iu49vC/Staging-Sites?t=dQlSC9S3VggdJP5w-0

cc @jasmussen


jasmussen:
Nice work — icons in the right place, npm run build done appropriately, all files committed. In principle this looks great! The minified SVG files testing well in Figma:
Screenshot 2023-03-16 at 09 56 04

I'm seeing some weird misalignment of the PDF version, not sure why. I'm also not sure I'm testing that right, or what you could've done differently here, so it's _probably fine?_ But if someone needs the PDF version, we can come back with a new PR and explore. The approach for it should've worked, so I'm not seeing it as a blocker.

Be aware I wasn't able to test the react component. But this seems good to go.

jasmussen (approved):

sejas (approved):

PR #326, opened 2023-03-29
DRAFT | Add New Copy Icon for Launchpad

sharpirate:
### Summary
I am proposing a new copy icon which will be used inside Launchpad:

gridicons-copy 2

### Reference
This is how the new copy icon (on the right) looks like in comparison to the existing clipboard icon (on the left):

new copy icon

cc @nuriapenya


Addison-Stavlo:
I see this is marked as a "Draft" are there any changes still pending here? any relevant testing instructions?

nuriapenya:
It seems that Gridicons is a library not used anymore in design. The idea moving forward will be to reuse the Gutenberg icons: https://wordpress.github.io/gutenberg/?path=/story/icons-icon--library and, eventually, update everything across the product. 

With that in mind, we should be using the `copy` icon, which looks similar:

Screenshot 2023-03-29 at 17 58 24


I would recommend size: **20x20px**

Figma: 9wpoYJrMlYKAKUYvVBMN74-fi-2957-22007

PR #338, opened 2026-07-07
Add React 19 support

simison:
- Bump React in `peerDependencies` to 19.
- Bump package version v3.4.2 → v3.4.3

Gridicons use stable React APIs (`React.createElement`) and plain props defaults, which are compatible across old and new React.

Ran a simple test with AI just in case:

---

I ran a version matrix in a temp harness with:
- `react`/`react-dom` `15.3.0`
- `16.14.0`
- `17.0.2`
- `18.3.1`
- `19.2.0`

For each version, I rendered both:
- an individual icon component (`dist/add.js`)
- the main `Gridicon` export (`dist/index.js`)

using `react-dom/server.renderToStaticMarkup`, and verified SVG output.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 _The versioning refers to the React component build._
 
-### Unreleased
+### v3.4.3 (2026-07-07)
 * React 19 support
 
 ### v3.4.2 (2023-03-16)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 _The versioning refers to the React component build._
 
+### Unreleased
+* React 19 support
+
 ### v3.4.2 (2023-03-16)
 * Icon added: "science"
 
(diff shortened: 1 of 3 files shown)


anomiex (approved):
Looks reasonable to me.

aduth (approved):
I was concerned about the use of `defaultProps` in `sources/react` templates, but apparently that's [still supported for class components](https://react.dev/blog/2024/04/25/react-19-upgrade-guide#removed-proptypes-and-defaultprops) so we should be okay. `propTypes` won't do anything in React 19 but also won't harm anything.

simison:
Thanks for reviews! Yeah, I don't wanna start removing unharmful stuff either, since it's kinda beautiful how an old package supports React v15–19.

anomiex:
Don't forget to [publish the release to npm](https://github.com/Automattic/gridicons#publishing-to-npm)!

simison:
@anomiex done! Had to get access first.

